PR #1, opened 2015-08-18
Clarification around UseAutofacMiddleware

eashi:
Before this change the documentation didn't make it clear that UseAutofacMiddleware also adds the registered Middleware of the container to the IAppBuilder, this change should clarify that

### DIFF
--- a/docs/integration/owin.rst
+++ b/docs/integration/owin.rst
@@ -11,8 +11,11 @@ Due to the differences in the way OWIN handles the application pipeline (detecti
 To take advantage of Autofac in your OWIN pipeline:
 
 * Reference the ``Autofac.Owin`` package from NuGet.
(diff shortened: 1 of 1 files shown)


tillig:
Ah, I see - we didn't actually touch on the fact that you can register middleware that will be injected by Autofac. Good catch. Let me take your changes and add a little bit, then update all at once.

tillig:
Hopefully I addressed the concern with [this explanation with a bit more discussion on how to get DI into your middleware.](https://github.com/autofac/Documentation/commit/caf8c03b2909e588001e91acb9778476bf2bd6b2)

Let me know if this is confusing or needs additional expansion. Thanks for the help!

alexmg:
Looks good to me! :shipit:

PR #2, opened 2015-08-26
Don't use LimitType at all to resolve ILog

piers7:
Even after changing where LimitType comes from, there _still_ seem to be scenarios where the sample as-written passes the wrong type to the log factory delegate. However, I realized we don't need to do any of that anyway, since we can grab the _actual_ instance type from the ParameterInfo.


tillig:
Thanks!

PR #6, opened 2016-01-23
Correcting MatchingScopeLifetimeTags

Jawvig:
In the per-request-scope, correcting the MatchingScopeLifetimeTags entries so they
all refer to RequestLifetimeScopeTag and not to AutofacWebRequest, which is the
name of the value of the RequestLifetimeScopeTag value, as per MatchingScopeLifetimeTags.cs


tillig:
Nice catch. Thanks!

PR #7, opened 2016-02-28
Update signalr.rst

AkosLukacs:
Example adding HubPipelineModule to the OWIN hosted example. Finding the correct way to do this required some googling. The example is based on this https://github.com/SignalR/SignalR/issues/2226


tillig:
Thanks for the PR! Looks like a good addition.

PR #11, opened 2016-06-08
builder.Register<Portfolio> does not compile

GeertVL-zz:
So instead I changed the documentation so it compiles correctly.


alexmg:
Thanks for the PR @GeertVL. Good catch.

PR #14, opened 2016-06-10
Correcting the metadata keys according to the registrations

GeertVL-zz:


tillig:
I cleaned this page up to use all "SendAllowed" so it's consistent with the attribute usage.

PR #15, opened 2016-06-10
The same in the attributes with send-allowed

GeertVL-zz:


tillig:
Looking at how Attribute Metdata works, it uses the name of the property, like PascalCase ("SendAllowed") so if we want to unify the example, it needs to use "SendAllowed." But, yeah, there are doc errors here. Let me fix 'em up. Thanks for the catch!

PR #21, opened 2016-10-06
Corrected to DependsByProp in the registration

Airn5475:
Previously: `cb.RegisterType<DependsByProperty>()`


tillig:
Nice catch, thanks!

PR #23, opened 2016-10-26
Updated NuGet package reference

MichaelSL:
The old package doesn't work with versions after 4.0.0. I hope this fix will save time to somebody.


tillig:
Great catch, thanks!

PR #24, opened 2016-11-09
Add section about pitfalls using single instance

angularsen:
Add section about pitfalls using single instance, linking to a github issue.


tillig:
Thanks for the PR! What I actually ended up doing is [splitting off a whole separate page](http://autofac.readthedocs.io/en/latest/lifetime/captive-dependencies.html) for it since it's not just a singleton vs. non-singleton issue and I wanted to expand on it to make it clear.

angularsen:
Great, that was a very nice article and the warning note should help people discover it.

PR #28, opened 2016-12-16
Incorrect scope variable used in example code

mikesigs:
The example accidentally uses `scope1` which was from the example above. Should be `scope3`.


tillig:
Oops, yup, thanks!

PR #33, opened 2017-01-23
Update index.rst

Afshintm:
The wording is wrong and new learner gets confused. It should be "you add components that implement services to the container" instead of "you add components to the container that implement services".


tillig:
I think both are correct English and I'm not sure the change is required. Thanks for the suggestion, but I think we'll pass on this

PR #34, opened 2017-01-29
Changed Register to RegisterType method

VanVector:


VanVector:
There is no method with signature like .Register<OnlineState>(). I assume there should be .RegisterType<OnlineState>().

alexmg:
Thanks @VanVector. These should indeed be `RegisterType`.

PR #35, opened 2017-02-06
WithKey is obsolete.

mikesigs:
After installing the `Autofac.Extras.AttributeMetadata` package I discovered the `WithKey` attribute is obsolete and you are pointed to `Use the Autofac.Features.AttributeFilters.KeyFilterAttribute from the core Autofac library instead.`


tillig:
Nice catch, thanks!

PR #36, opened 2017-02-22
Fix formatting

gldraphael:


tillig:
Nice catch, thanks!

gldraphael:
The line after the bullet points reads: "In this case, the component is SomeType and the services it exposes are SomeType and IService."

Does that belong to point 2?

tillig:
Nope.

PR #39, opened 2017-03-22
Added info for JSON and XML config assemblies

nCubed:
The updates to Microsoft.Extensions.Configuration separates JSON and XML assemblies. This update makes it clear where to get the assemblies.

### DIFF
--- a/docs/configuration/xml.rst
+++ b/docs/configuration/xml.rst
@@ -27,7 +27,7 @@ The basic steps to getting configuration set up with your application are:
 
 1. Set up your configuration in JSON or XML files that can be read by ``Microsoft.Extensions.Configuration``.
   * JSON Configuration uses ``Microsoft.Extensions.Configuration.Json``
-  * XML Configuration uses ``Microsoft.Extensions.Configuration.XML``
+  * XML Configuration uses ``Microsoft.Extensions.Configuration.Xml``
 2. Build the configuration using the ``Microsoft.Extensions.Configuration.ConfigurationBuilder``.
 3. Create a new ``Autofac.Configuration.ConfigurationModule`` and pass the built ``Microsoft.Extensions.Configuration.IConfiguration`` into it.
 4. Register the ``Autofac.Configuration.ConfigurationModule`` with your container.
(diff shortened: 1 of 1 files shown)


docs/configuration/xml.rst
```diff
@@ -26,6 +26,8 @@ Quick Start
 The basic steps to getting configuration set up with your application are:
 
 1. Set up your configuration in JSON or XML files that can be read by ``Microsoft.Extensions.Configuration``.
+  * JSON Configuration uses ``Microsoft.Extensions.Configuration.Json``
+  * XML Configuration uses ``Microsoft.Extensions.Configuration.XML``
```
tillig: It should be `Microsoft.Extensions.Configuration.Xml` (the "XML" in the library is cased like "Xml" for the NuGet package).

tillig: "Configuration" should be lower case - "JSON configuration uses..." and "XML configuration uses..."

docs/configuration/xml.rst
```diff
@@ -82,6 +84,8 @@ Build up your configuration and register it with the Autofac ``ContainerBuilder`
 
     // Add the configuration to the ConfigurationBuilder.
     var config = new ConfigurationBuilder();
+    // config.AddJsonFile comes from Microsoft.Extensions.Configuration.Json
+    // config.AddXmlFile comes from Microsoft.Extensions.Configuration.XML
```
tillig: `Microsoft.Extensions.Configuration.Xml`

nCubed:
I've updated the casing. This should match the nuget package names now for both Json and Xml.

docs/configuration/xml.rst
```diff
@@ -27,7 +27,7 @@ The basic steps to getting configuration set up with your application are:
 
 1. Set up your configuration in JSON or XML files that can be read by ``Microsoft.Extensions.Configuration``.
   * JSON Configuration uses ``Microsoft.Extensions.Configuration.Json``
-  * XML Configuration uses ``Microsoft.Extensions.Configuration.XML``
+  * XML Configuration uses ``Microsoft.Extensions.Configuration.Xml``
```
tillig: Still need "configuration" in the sentence ("XML configuration" and "JSON configuration") to be lower case (not "JSON Configuration" - "Configuration" isn't a proper noun).

nCubed: Good catch!

nCubed:
If I have another pull request, I'll be more cognizant of grammar and structure. I believe we should be good to go now.

tillig:
Thanks very much for the contribution! Sorry to seem pedantic, and very much appreciate your patience.

nCubed:
TBH, I appreciate you being so pedantic; I'm the same way. I think it takes more patience on your side than mine to ensure we contribute to the project in a manner consistent with the project owner's vision and style.

PR #45, opened 2017-10-23
Update aspnetcore.rst

divinci:
I propose the change based on this thinking.

Line 361 - There is no this.ApplicationContainer in a common Startup.cs class.
Line 363 - The MultitenantContainer constructor parameter 'container' does not exist.

So changing line 361 to possibly fix.


tillig:
Looks good - thanks!

PR #51, opened 2018-02-24
Note added to Mulitenant config section

charlvictor:
Added a note about the ConfigureServices method signature that needs to be changed to enable returning the Autofac service provider. This tripped me up a little because by default its void. Thought it might help newcomers to point to the documentation.

### DIFF
--- a/docs/integration/aspnetcore.rst
+++ b/docs/integration/aspnetcore.rst
@@ -368,6 +368,11 @@ Here's an example of what you do in ``Program.Main``:
       // This is what the middleware will use to create your request lifetime scope.
       public static MultitenantContainer ApplicationContainer { get; set; }
     }
+    
+.. note::
+    By default ASP.NET Core has a void return type for method ``ConfigureServices(IServiceCollection services)``. This needs to be      
+    updated to return the ``IServiceProvider`` type as shown above. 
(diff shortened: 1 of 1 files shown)


tillig:
Did you miss the third bullet (one of the three things you have to do to enable multitenant support) that read "In your `Startup.ConfigureServices` method create your multitenant container and return an `AutofacServiceProvider` using that container." ? Would it help if I added something right to that bullet to point out you can't return a service provider with a void method? We do show a fully working example... I don't really want to put an additional, redundant note to that effect at the end; we try to balance the "documentation completeness" with "too long; didn't read" and we get hit a lot with TL;DR. I'd prefer to just update the items above the samples.

charlvictor:
@tillig I read through the section again, and while I see the example code, it is still not apparent that I am allowed to change the method signature from void to return an IServiceProvider object. I am new to ASP.NET Core, so I did not know that you could change the method signature.  Just thought that if there is just some note there it would help understand the sample code better.

tillig:
I switched the third bullet to be

> Change your `Startup.ConfigureServices` method to return `IServiceProvider`, create your multitenant container, and return an `AutofacServiceProvider` using that container.

That should both remove the ambiguity and keep the docs somewhat concise. If it isn't clear from that, I'm not sure there's much more we can do without venturing into TL;DR territory because all through that page there are different ways to set up the container and it switches from `void` to `IServiceProvider` on `ConfigureServices`. We can't call it out every time; folks will need an attention to detail in setting things up. There are also plenty of docs about that difference on out in the wild.

Hopefully this will help.

PR #52, opened 2018-04-09
Updated notes for IStartable ordering to reflect new code changes

babrekel:
@tillig Would love some feedback on this one. I added a code example which is kind of lengthy -- I can change it to use expression bodied members, but it might look counter intuitive to the reader, I placed the example at the bottom of this comment.

Also, I wanted to remove the guidance of not allowing other exports from `IStartable` components, since it might be able to solve certain scenarios quite well. One pattern that we frequently see is this:

    class PrefetchingCache : IStartable, ISomeKindOfCache
    {
        private Task foreverRunningTask;

        public void Start()
        {
            // initialize recurring timer and fetch data for the first time.
            foreverRunningTask = RunAsync();
        }

        public async Task RunAsync()
        {
            while (!shuttingdown)
            {
                this.data = await FetchChanges();
                await Task.Delay(...);
            }
        }

        // ISomeKindOfCache implementation, that uses this.data
    }

Please let me know if this makes sense, if you feel differently, I'm of course happy to undo the change :) 

Should the "long contrived" unit test example at the bottom of the page be modified? I'm not sure if it's there to clarify an alternative to the `IStartable` issue, or just to explain build callbacks more clearly.

Shorter version of code sample as per comment above:

    static void Main()
    {
        builder.RegisterType<Startable1>().AsSelf().As<IStartable>().SingleInstance();
        builder.RegisterType<Startable2>().As<IStartable>().SingleInstance();
        builder.Build();
    }

    class Startable1 : IStartable
    {
        public Startable1() => Console.WriteLine("Startable1 activated");
        public void Start() => Console.WriteLine("Startable1 started");
    }

    class Startable2 : IStartable
    {
        public Startable2() => Console.WriteLine("Startable1 activated");
        public void Start() => Console.WriteLine("Startable1 started");
    }

### DIFF
--- a/docs/lifetime/startup.rst
+++ b/docs/lifetime/startup.rst
@@ -18,7 +18,7 @@ A **startable component** is one that is activated by the container when the con
 
 The key is to implement the ``Autofac.IStartable`` interface. When the container is built, the component will be activated and the ``IStartable.Start()`` method will be called.
 
-**This only happens once, for a single instance of each component, the first time the container is built.** Resolving startable components by hand won't result in their ``Start()`` method being called. It isn't recommended that startable components implement other services, or be registered as anything other than ``SingleInstance()``.
+**This only happens once, for a single instance of each component, the first time the container is built.** Resolving startable components by hand won't result in their ``Start()`` method being called. It isn't recommended that startable components are registered as anything other than ``SingleInstance()``.
 
 Components that need to have something like a ``Start()`` method called *each time they are activated* should use :doc:`a lifetime event <events>` like ``OnActivated`` instead.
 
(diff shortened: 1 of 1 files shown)


tillig:
It may be OK to remove the caution about `IStartable` not implementing other services, though I haven't done an exhaustive check. Since we didn't previously have the feature to run `IStartable` in dependency order it was possible that by exposing multiple services one `IStartable` could start but not yet have allowed a different `IStartable` dependency to do its `Start()` first. Let's just go with it and see what happens.

The contrived unit test both serves as a demo for callbacks and to show an alternative to `IStartable`. We have a lot of folks asking how to avoid references to Autofac from component libraries so if someone defines their own interface that acts as `IStartable` this is a concrete example that shows how build callbacks can be used to aid that implementation.

I think the changes here look great. I'm going to hold on merging the PR until a release with the updated `IStartable` fix is out so we can avoid confusion.

Thanks!

PR #53, opened 2018-04-18
Use lifetime scope rather than container

tblong:
This is an update to show that the lifetime scope created should be used instead of the container directly.


tillig:
Whoops! Good catch. Thanks!

PR #55, opened 2018-06-04
Update autofake docs

blairconrad:
I'm happy to squash if you don't like the separate commits. Or (as always) to make any other changes you'd like. No nitpick is too nitty!


tillig:
Looks good, thanks!

PR #60, opened 2018-12-06
Castle.Core Versionning Issue Resolution

UnLuckyNiko:
Propose a way to circumvent Castle.Core versionning issues using AutoFac


tillig:
I see what you're saying here, but what you're doing totally avoids using the Autofac.DynamicProxy stuff at all. The troubleshooting we're trying to offer in the docs is how you can keep using the Autofac.DynamicProxy library when you run into the version issue.

UnLuckyNiko:
Hmm I get it, Sorry I didn't realize I was completely bypassing that. Nevertheless, I have yet to find anything that helps me with the original issue as the 2 resolutions on the website didn't resolve it.

PR #63, opened 2019-01-18
Register method should be RegisterType<>

reyou:


tillig:
Whoops! Yep, I think that's left over from the 2.x days. Good catch!

Issue #69, opened 2019-03-02
Add Preparing event documentation


tillig:
The [lifetime event docs](https://autofac.readthedocs.io/en/latest/lifetime/events.html) have OnActivating, OnActivated but not [OnPreparing](https://github.com/autofac/Autofac/blob/6dde84e5b0a3f82136a0567a84da498b04e1fa2d/src/Autofac/Core/IComponentRegistration.cs#L83).

alistairjevans:
Taking a brief look at this, it looks like the comments on [OnPreparing](https://github.com/autofac/Autofac/blob/6dde84e5b0a3f82136a0567a84da498b04e1fa2d/src/Autofac/Core/IComponentRegistration.cs#L83) you linked to may not be correct, because there is no Instance property on the [PreparingEventArgs](https://github.com/autofac/Autofac/blob/6dde84e5b0a3f82136a0567a84da498b04e1fa2d/src/Autofac/Core/PreparingEventArgs.cs) class.

If I'm drafting some docs for OnPreparing shall I raise and fix an issue on Autofac to update the comment?

tillig:
Yup, that works.

PR #75, opened 2019-07-16
Added OnPreparing Docs

alistairjevans:
I've drafted some basic OnPreparing Docs, to address issue #69. You'll have to forgive me if the content is way off base, the only use-case I could find was in the Dynamic Proxy support.  I struggled to think of a non-proxy-related use-case that can't be met by parameter registration, so I put a suggestion in there to that effect.


tillig (approved):
This is great. I don't really have any use cases around when exactly Preparing would be used, either. We can always add one later if we think of it.

PR #77, opened 2019-07-31
Updated Link regarding Asp.NET Core Integration

ssimontis:
The link to StartupLoader class was pointing to an archived product on GitHub; the functionality has been integrated into the Aspnetcore project, so I tried to hunt down the exact code. StartupLoader no longer exists, but I located several extension methods which give developers control over the startup process by providing an opportunity to override the convention-based initialization process.

### DIFF
--- a/docs/integration/aspnetcore.rst
+++ b/docs/integration/aspnetcore.rst
@@ -235,7 +235,7 @@ This means you don't necessarily have to use :doc:`Autofac configuration <../con
       }
     }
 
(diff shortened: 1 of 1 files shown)


alistairjevans (changes requested):
I agree the link needs changing, but I think the extensions you've linked to actually control the WebHost builder in the Program.cs, rather than how the startup class behaves.
I think the StartupLoader is still a thing anyway, perhaps the link just needs updating to:
https://github.com/aspnet/AspNetCore/blob/f3f9a1cdbcd06b298035b523732b9f45b1408461/src/Hosting/Hosting/src/Internal/StartupLoader.cs
?

tillig:
Agreed, the StartupLoader is still a thing. The link may need to be updated, but not removed. If an additional link needs to be added for other extension methods and/or clarity, that's fine, too, but it's not mutually exclusive.

PR #78, opened 2019-08-18
Updated the WebAPI Filter documentation following the release of 4.3.0

alistairjevans:
Added a section on continuation-style filters, and mostly rewrote the section on registering filters to go over the new methods and the ability to chain them together.

### DIFF
--- a/docs/integration/webapi.rst
+++ b/docs/integration/webapi.rst
@@ -113,12 +113,19 @@ After building your container pass it into a new instance of the ``AutofacWebApi
 
 Provide Filters via Dependency Injection
 ========================================
-Because attributes are created via the reflection API you don't get to call the constructor yourself. That leaves you with no other option except for property injection when working with attributes. The Autofac integration with Web API provides a mechanism that allows you to create classes that implement the filter interfaces (``IAutofacActionFilter``, ``IAutofacAuthorizationFilter`` and ``IAutofacExceptionFilter``) and wire them up to the desired controller or action method using the registration syntax on the container builder.
+Because attributes are created via the reflection API you don't get to call the constructor yourself. 
+That leaves you with no other option except for property injection when working with attributes. 
+The Autofac integration with Web API provides a mechanism that allows you to create classes that 
+implement the filter interfaces 
+(``IAutofacActionFilter``, ``IAutofacContinuationActionFilter``, ``IAutofacAuthorizationFilter`` and ``IAutofacExceptionFilter``) 
+and wire them up to the desired controller or action method using the registration syntax on the container builder.
 
 Register the Filter Provider
 ----------------------------
 
-You need to register the Autofac filter provider implementation because it does the work of wiring up the filter based on the registration. This is done by calling the ``RegisterWebApiFilterProvider`` method on the container builder and providing an ``HttpConfiguration`` instance.
+You need to register the Autofac filter provider implementation because it does the work of wiring up the filter 
+based on the registration. This is done by calling the ``RegisterWebApiFilterProvider`` method on the container 
+builder and providing an ``HttpConfiguration`` instance.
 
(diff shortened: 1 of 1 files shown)


docs/integration/webapi.rst
```diff
@@ -128,7 +135,9 @@ You need to register the Autofac filter provider implementation because it does
 Implement the Filter Interface
 ------------------------------
 
-Instead of deriving from one of the existing Web API filter attributes your class implements the appropriate filter interface defined in the integration. The filter below is an action filter and  implements ``IAutofacActionFilter`` instead of ``System.Web.Http.Filters.IActionFilter``.
+Instead of deriving from one of the existing Web API filter attributes your class implements the
```
tillig, line 138: If there's a header for "Continuation Action Filter Interface" we should probably have one at the same level (`*****`) for something like "Standard Action Filter Interface." The first sentence here ("Instead of deriving...in the integration.") would stay under the main "Implement the filter interface" header, but then folks could see examples of the two action filters under their respective headers.

docs/integration/webapi.rst
```diff
@@ -154,79 +163,136 @@ Instead of deriving from one of the existing Web API filter attributes your clas
...
-- ``AsWebApiActionFilterFor<TController>()``
-- ``AsWebApiActionFilterOverrideFor<TController>()``
-- ``AsWebApiAuthorizationFilterFor<TController>()``
-- ``AsWebApiAuthorizationOverrideFilterFor<TController>()``
-- ``AsWebApiAuthenticationFilterFor<TController>()``
-- ``AsWebApiAuthenticationOverrideFilterFor<TController>()``
-- ``AsWebApiExceptionFilterFor<TController>()``
-- ``AsWebApiExceptionOverrideFilterFor<TController>()``
+``AsWebApi{FilterType}ForAllControllers``
+  Register this filter to run for all action methods on all controllers, in the same way as registering a global Web API filter.
```
tillig, line 226: This is a good refactoring of the doc, but on initial read it's not clear what options there are for `{FilterType}`. I do see the list down below, but maybe it would be more clear to move it up here where I see the first occurrence of `{FilterType}`; or put something in here that says "you can see the list of supported filter types below." Something to help folks connect the stuff up here to the stuff down there.

tillig (changes requested):

docs/integration/webapi.rst
alistairjevans: I've added that heading and some extra words to preamble the standard interface.

docs/integration/webapi.rst
alistairjevans: I've moved the filter types up to the top of the section; I expect users don't want to have to scroll down first, then scroll up again.

tillig (approved):
LGTM. Thanks!

alistairjevans:
Does anything need doing to trigger a build of the docs? Just that the last docs build according to RTD (https://readthedocs.org/projects/autofac/?badge=latest) was 5 weeks ago, but there's now been 2 PRs merged since then. Isn't there meant to be a web hook?

tillig:
There is a web hook, but ReadTheDocs gets... hung up sometimes. Let me go look.

tillig:
I re-saved the configuration for the RTD build and it seems to have triggered the build. I guess we'll have to see if it fixes it. [Here's the list of builds for reference.](https://readthedocs.org/projects/autofac/builds/)

alistairjevans:
OK, thanks Travis.  I'll keep an eye on subsequent PRs.

tillig:
I figured out the build thing. Somehow _all the webhook connections disappeared_ between RTD and GitHub. I had some other projects also using RTD, also not building. I went back through the initial setup steps to verify how things were working and... webhooks lost. I've recreated them so hopefully it won't be a problem again.

PR #80, opened 2019-09-08
Add required service-collection extension method to description for multi-tenancy in ASP.NET Core

alsami:
This adds the missing line that needs to be called in `ConfigureServices(IServiceCollection services)` to register the multitenant-middleware in ASP.NET Core.


tillig (approved):
LGTM

PR #81, opened 2019-09-08
fix: add missing language declaration in webapi.rst to fix make html

alsami:
This adds `csharp` to a source-code declaration in `webapi.rst` to make `make` work again.


tillig (approved):
LGTM

PR #82, opened 2019-09-11
Added blazor page. Tidied main index description.

johneking:
I've added a new page with basic instructions covering the complications for client-side DI in Blazor as per #1025. It should potentially be part of the ASP.NET Core page but that seems to be getting a little bloated - maybe it's worth considering splitting that down soon, similar to the ASP.NET section?

I also noticed the somewhat outdated description for general Autofac target frameworks on the main index page, which looks a little odd, so I've snuck in a small update for that here too.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -3,7 +3,7 @@ Welcome to Autofac's documentation!
 
 .. image:: logo.png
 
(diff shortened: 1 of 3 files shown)


tillig (approved):
LGTM

tillig:
You're right, the .NET Core stuff probably needs to be adjusted. There's "generic app hosting" now rather than just ASP.NET Core app hosting, there's the Blazor stuff you added which sort of touches on that, there are a few ASP.NET Core specific things... I'll add an issue here to track that. Good idea!

Issue #85, opened 2019-09-25
What is the "Without ConfigureContainer" mechanism?


jimmcslim:
The page https://github.com/autofac/Documentation/blob/master/docs/integration/aspnetcore.rst has a comment in a code example:

````
  // If you need a reference to the container, you need to use the
  // "Without ConfigureContainer" mechanism shown later.
````

However it is not clear if this is actually shown?

Also, is it possible to combine `ConfigureContainer` and `AddControllersAsServices()`... an example for that would be helpful as well.

**Comment by alsami** [comment]
Ups, something got lost there. Which version of ASP.NET-Core are you on?

Anyway, when **below** ASP.NET Core 3. You can do the following in your `Startup.cs`.

```csharp
public IServiceProvider ConfigureServices(IServiceCollection services)
{
   // add stuff to services, like services.AddMvc();
   var containerBuilder = new ContainerBuilder();
   // add stuff to the containerBuilder
   containerBuilder.Populate(services);
   var container = container.Build();
   return new AutofacServiceProvider(container);
}
```

I will create a PR soon to fix that. 



> Also, is it possible to combine `ConfigureContainer` and `AddControllersAsServices()`... an example for that would be helpful as well.

What do you mean by that? Would you eloberate?

jimmcslim:
Actually I don't think I had a further problem, or whatever it was, I've forgotten it! Thanks!

alsami:
Okay, then your question should be answered now. We have added a sample again on how to do it with ASPNET Core version < 3.0 and without configure container.

Issue #83, opened 2019-09-11
Break down .NET Core / ASP.NET Core docs


tillig:
Since ASP.NET Core first came out, there's a lot that has changed. It would probably be good to reorganize the documentation around .NET Core things. There's a lot to discuss:

- .NET Core - basic usage of the `AutofacServiceProvider` when not used in any sort of hosted app. Think console app or something that doesn't have the hosting bits wrapped up in it.
- .NET Core generic hosting - new in .NET Core 3, this creates a concept of generic app startup for an application. Affects how ASP.NET Core handles app hosting.
- ASP.NET Core - in .NET Core 1.0 - 2.2 this is all about web hosting; in .NET Core 3.0 it's an add-on for generic hosting. There are also things like controllers-as-services, differences between .NET Core and .NET Desktop (WebAPI, MVC), that sort of thing. Even just in this topic there's a lot to break down.
- Blazor - Similar to ASP.NET Core in that it uses a hosting model but may need to have some information about dealing with optimization (e.g., how the [Xamarin and .NET Native linkers filter out types they see as unused](https://autofaccn.readthedocs.io/en/latest/advanced/cross-platform-apps.html)... does Blazor do that?)
- Multitenant support - we have a multitenant service provider factory. Need to keep examples for that stuff.

**This isn't a suggestion that each of these topics get the pages I just listed.** That's probably not the right granularity. For example, it may be better to put multitenant notes on specific pages with examples rather than on its own page. Or maybe the .NET Core and generic hosting stuff all go on one page. **Part of the task here is to come up with a good breakdown/structure for this that can be extended** as time goes on and new app types arise.

tillig:
Even more has changed here and .NET Core is becoming .NET. I can't think of a better way to break down these docs and get people the info they need. I somewhat think we need to have more info about what's deprecated/in maintenance mode, and possibly move some of the deprecated docs out to the README in the repo and remove them from the body of the site. For example, the DomainServices support with Silverlight hasn't been a thing for _years_ but we still have docs for it.

PR #86, opened 2019-09-25
Re-add sample for returning an service-provider from configure-service delegate

alsami:
This PR ...

* Re-adds the old fashioned way of using a custom service-provider for ASP.NET-Core, where you would return an `IServiceProvider` from `ConfigureService(IServiceCollection services)` delegate.
* Changes the version range for `sp.AddAutofac()` because it does work for 2.1 and 2.2 of ASP.NET Core as well.

Relates to #85 and probably even a bit to #83


tillig (approved):
LGTM

PR #92, opened 2019-12-02
fix: clarify documentation for AutofacChildLifetimeScopeServiceProviderFactory

### LINKED ISSUE #90: AutofacChildLifetimeScopeServiceProviderFactory setup documentation is invalid.
Incorrect documentation page:
https://autofaccn.readthedocs.io/en/latest/integration/aspnetcore.html#using-a-child-scope-as-a-root

The documentation explicitly showing how to setup the `AutofacChildLifetimeScopeServiceProviderFactory` where it takes 0 constructor arguments is incorrect. 

```
    var host = Host.CreateDefaultBuilder(args)
        .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory())
        .ConfigureWebHostDefaults(webHostBuilder => {
```

As of `Autofac.Extensions.DependencyInjection, Version=5.0.1.0` the `AutofacChildLifetimeScopeServiceProviderFactory` does not have any empty constructor.

https://github.com/autofac/Autofac.Extensions.DependencyInjection/blob/master/src/Autofac.Extensions.DependencyInjection/AutofacChildLifetimeScopeServiceProviderFactory.cs

alsami:
This PR clarifies the usage of `AutofacChildLifetimeScopeServiceProviderFactory`.

We might want to add some samples in the examples repo, if they don't exist. I'll create a PR hopefully soon.

fixes #90

### DIFF
--- a/docs/integration/aspnetcore.rst
+++ b/docs/integration/aspnetcore.rst
@@ -399,21 +399,51 @@ In standard ASP.NET Core integration and generic hosted application support ther
 
     public class Program
     {
-      public static void Main(string[] args)
+      public static async Task Main(string[] args)
       {
+        // create the root-container and register global dependencies
+        var containerBuilder = new ContainerBuilder();
+
+        builder.RegisterType<SomeGlobalDependency>()
+          .As<ISomeGlobalDependency>()
+          .InstancePerLifetimeScope();
+
+        var container = containerBuilder.Build();
+
         // The UseServiceProviderFactory call attaches the
         // Autofac provider to the generic hosting mechanism.
-        var host = Host.CreateDefaultBuilder(args)
-            .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory())
+          var hostOne = Host
+            .CreateDefaultBuilder(args)
+            .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container.BeginLifetimeScope("root-one")))
             .ConfigureWebHostDefaults(webHostBuilder => {
               webHostBuilder
-                .UseContentRoot(Directory.GetCurrentDirectory())
-                .UseIISIntegration()
-                .UseStartup<Startup>();
+                .UseContentRoot(AppContext.BaseDirectory)
+                // Each host listens to a different URL, they have the same root container to share SingleInstance
+                // things, but they each have  their own logical root lifetime scope. Registering things
+                // as `InstancePerMatchingLifetimeScope("root-one")` (the name of the scope given above)
+                // will result in a singleton that's ONLY used by this first host.
+                .UseUrls("http://localhost:5000")
+                .UseStartup<StartupOne>();
             })
             .Build();
 
-        host.Run();
+        // The UseServiceProviderFactory call attaches the
+        // Autofac provider to the generic hosting mechanism.
+          var hostTwo = Host
+            .CreateDefaultBuilder(args)
+            .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container.BeginLifetimeScope("root-two")))
+            .ConfigureWebHostDefaults(webHostBuilder => {
+              webHostBuilder
+                .UseContentRoot(AppContext.BaseDirectory)
+                // As with the first host, the second host will share the root container but have its own
+                // root lifetime scope `root-two`. Things registered `InstancePerMatchingLifetimeScope("root-two")`
+                // will be singletons ONLY used by this second host.
+                .UseUrls("http://localhost:5001")
+                .UseStartup<StartupTwo>();
+            })
+            .Build();
+
+        await Task.WhenAll(hostOne.RunAsync(), hostTwo.RunAsync())
       }
     }
 
@@ -421,9 +451,10 @@ This will change how your ``Startup`` class works - you won't use a ``ContainerB
 
 .. sourcecode:: csharp
 
-    public class Startup
+    public class StartupOne
     {
-      public Startup(IHostingEnvironment env)
+      // IHostingEnvironment when running applications below ASP.NET Core 3.0
+      public Startup(IWebHostEnvironment env)
       {
         // Fill this in if needed...
       }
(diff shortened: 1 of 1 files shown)


tillig (changes requested):
The example definitely needs to be updated, but the value of the `AutofacChildLifetimeScopeServiceProivderFactory`, I think, is more about the ability to give it a child lifetime scope _of your own creation_, like if you have an existing container and you created your own root lifetime scope from that. The idea was from folks who are trying to self-host multiple app types in a single application - a shared root container, but each app type can have a separate logical root scope.

```c#
public class Program
{
  public static void Main(string[] args)
  {
    builder.RegisterType<SomethingGlobal>();

    // First host has a named "root lifetime scope" of "host1".
    var host1 = Host.CreateDefaultBuilder(args)
      .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container.BeginLifetimeScope("host1")))
      .ConfigureWebHostDefaults(webHostBuilder => {
        webHostBuilder
          .UseContentRoot(Directory.GetCurrentDirectory())
          .UseStartup<Startup1>();
          })
        .Build()
        .RunAsync();

    // Second host has a named "root lifetime scope" of "host2".
    var host2 = Host.CreateDefaultBuilder(args)
      .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container.BeginLifetimeScope("host2")))
      .ConfigureWebHostDefaults(webHostBuilder => {
        webHostBuilder
          .UseContentRoot(Directory.GetCurrentDirectory())
          .UseStartup<Startup2>();
          })
        .Build()
        .RunAsync();

    // Each host listens on different ports, does different things, and shares a root
    // container but has specific registrations in their own lifetime scopes.
    Task.WaitAll(host1, host2);
  }
}
```

While the updated example shows creation of a container, it doesn't really show folks why that's any different or better than just creating the container directly in `Startup`.

alsami:
Hey @tillig 

sorry that it took me a while to update the PR been rather busy at work lately. I hope the sample is more satisfying now. I will add a proper code sample to the examples repository and raise a new PR for that.

Merry christmas!

docs/integration/aspnetcore.rst
```diff
@@ -399,31 +399,85 @@ In standard ASP.NET Core integration and generic hosted application support ther
...
+
+        var container = containerBuilder.Build();
+
         // The UseServiceProviderFactory call attaches the
         // Autofac provider to the generic hosting mechanism.
-        var host = Host.CreateDefaultBuilder(args)
-            .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory())
+          var hostOne = Host
+            .CreateDefaultBuilder(args)
+            .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container))
```
tillig:
```suggestion
            .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container.BeginLifetimeScope("root-one")))
```

The idea here is that the "root" can be a child scope of the container, it doesn't have to be a container at all. I'll update the two examples with some suggestions and we should be able to call it good.

docs/integration/aspnetcore.rst
```diff
@@ -399,31 +399,85 @@ In standard ASP.NET Core integration and generic hosted application support ther
...
             .ConfigureWebHostDefaults(webHostBuilder => {
               webHostBuilder
-                .UseContentRoot(Directory.GetCurrentDirectory())
-                .UseIISIntegration()
-                .UseStartup<Startup>();
+                .UseContentRoot(AppContext.BaseDirectory)
+                // each host listens to a different URL while still using the same root-container
```
tillig:
```suggestion
                // Each host listens to a different URL, they have the same root container to share SingleInstance
                // things, but they each have  their own logical root lifetime scope. Registering things
                // as `InstancePerMatchingLifetimeScope("root-one")` (the name of the scope given above)
                // will result in a singleton that's ONLY used by this first host.
```

docs/integration/aspnetcore.rst
```diff
@@ -399,31 +399,85 @@ In standard ASP.NET Core integration and generic hosted application support ther
...
+                .UseStartup<StartupOne>();
             })
             .Build();
 
-        host.Run();
+        // The UseServiceProviderFactory call attaches the
+        // Autofac provider to the generic hosting mechanism.
+          var hostTwo = Host
+            .CreateDefaultBuilder(args)
+            .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container))
```
tillig:
```suggestion
            .UseServiceProviderFactory(new AutofacChildLifetimeScopeServiceProviderFactory(container.BeginLifetimeScope("root-two")))
```

docs/integration/aspnetcore.rst
```diff
@@ -399,31 +399,85 @@ In standard ASP.NET Core integration and generic hosted application support ther
...
+            .ConfigureWebHostDefaults(webHostBuilder => {
+              webHostBuilder
+                .UseContentRoot(AppContext.BaseDirectory)
+                // each host listens to a different URL while still using the same root-container
```
tillig:
```suggestion
                // As with the first host, the second host will share the root container but have its own
                // root lifetime scope `root-two`. Things registered `InstancePerMatchingLifetimeScope("root-two")`
                // will be singletons ONLY used by this second host.
```

tillig (changes requested):
I put a couple of update suggestions here to make it clear to readers that they can control the root lifetime scope - it doesn't have to be a container at all. Also a couple of notes to show how you can have a "singleton" that's only used by one of the two hosts, making a "logical root" that isn't the container.

tillig:
I think an even more advanced use case is, like...

- Main program creates a container and registers things.
- Main program creates a child lifetime scope (tagged) and registers more per-lifetime-scope things in there.
- Main program hands _that child lifetime scope_ to the AutofacChildLifetimeScopeServiceProviderFactory.
- The rest of the web app startup continues as normal using that factory.

This was more popular in Web API where there was, say, a Windows service that had a container and wanted to self-host an API endpoint but not sully the container with API-specific junk. Probably less popular now, but that's the idea.

alsami:
I think I have addressed all the suggestions you have made.

The advanced use-case you have described above sounds good. I will try to build something for like that for the examples repository.

tillig (approved):
Sweet, thanks!

PR #93, opened 2019-12-05
Update relationships.rst

Brondahl:
The documentation on Dynamic and Paramaterised Dynamic Instantiation isn't quite clear enough on how it interacts with the constructors you declared. Hopefully I've made it clearer, and not introduced any material misconceptions.

### DIFF
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -221,7 +221,8 @@ Example:
 
         public void M(P existingPea)
         {
-            var b = _newB(42, existingPea); //the needed Q and R will be resolved by Autofac, but not the P since the existingPea will be used.
+            // The Q and R will be resolved by Autofac, but P will be existingPea instead.
+            var b = _newB(42, existingPea);
             b.DoSomething();
         }
     }
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -230,7 +230,7 @@ Example:
 Internally, Autofac determines what values to use for the constructor args, solely based on the types, and behaves as though we've temporarily defined the input values as the objects that should be used to resolve those types. A direct consequence of this is that  **auto-generated function factories cannot have duplicate types in the input parameter list.** See below for further notes on this.
 
 
-**Lifetime scopes are respected** using this relationship type, just as they are when using delegate factories. If you register an object as ``InstancePerDependency()`` and call the ``Func<X, Y, B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<X, Y, B>`` to resolve the object more than once, you will get *the same object instance every time regardless of the different parameters you pass in.* Just passing different parameters will not break the respect for the lifetime scope.
+**Lifetime scopes are respected** using this relationship type, just as they are when using :doc:`delegate factories <../advanced/delegate-factories>`. If you register an object as ``InstancePerDependency()`` and call the ``Func<X, Y, B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<X, Y, B>`` to resolve the object more than once, you will get *the same object instance every time regardless of the different parameters you pass in.* Just passing different parameters will not break the respect for the lifetime scope.
 
 Parameterized Instantiation on types with duplicate Types in their constructor args.
 -------------------------------------------
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -176,7 +176,7 @@ An example of this relationship looks like:
 
 Parameterized Instantiation (Func<X, Y, B>)
 -------------------------------------------
-You can also use an *auto-generated factory* to provide parameters when creating an new instance of the object, where the constructor of the object calls for some additional parameters. As in the case of Dynamic Instantiation above, we're still doing something equivalent to calling ``.Resolve<B>()``, but in this case we're passing strongly-typed parameters to the resolution function. This is an alternative to :doc:`passing parameters during registration <../register/parameters>` or :doc:`passing during manual resolution <../resolve/parameters>`:
+You can also use an *auto-generated factory* to provide parameters when creating an new instance of the object, where the constructor of the object calls for some additional parameters. While the ``Func<B>`` relationship is similar to ``Resolve<B>()``, the ``Func<X, Y, B>`` relationship is like calling ``Resolve<B>(TypedParameter.From<X>(x), TypedParameter.From<Y>(y))`` - a resolve operation that has typed parameters. This is an alternative to :doc:`passing parameters during registration <../register/parameters>` or :doc:`passing during manual resolution <../resolve/parameters>`:
 
 .. sourcecode:: csharp
     public class B
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -154,7 +154,7 @@ This relationship is also useful in cases like :doc:`WCF integration <../integra
 
 Note that whilst this could be used to simply invoke a parameterless constructor of the target type ``B``, that isn't all it's limited to. As stated above, invoking the ``Func<B>`` is more akin to calling ``Resolve<B>``, and hence Autofac can wire up other types passed into the constructor, as long as it knows how to provide those types. Hence this calling format can be used for any type where all the ctor dependencies can be resolved by the container.
 
-Another corollary of this behaving like calling ``Resolve<B>()`` is that **Lifetime scopes are respected** using this relationship type. If you register an object as ``InstancePerDependency()`` and call the ``Func<B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<B>`` to resolve the object more than once, you will get *the same object instance every time*.
+Further, lifetime scopes are respected. If you register an object as ``InstancePerDependency()`` and call the ``Func<B>`` multiple times, you'll get a new instance each time; if you register an object as ``SingleInstance()`` and call the ``Func<B>`` to resolve the object more than once, you will get *the same object instance every time*.
 
 An example of this relationship looks like:
 
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -152,7 +152,7 @@ Using an *auto-generated factory* can let you resolve a new `B` programmatically
 
 This relationship is also useful in cases like :doc:`WCF integration <../integration/wcf>` where you need to create a new service proxy after faulting the channel.
 
-Note that whilst this could be used to simply invoke a parameterless constructor of the target type ``B``, that isn't all it's limited to. As stated above, invoking the ``Func<B>`` is more akin to calling ``Resolve<B>``, and hence Autofac can wire up other types passed into the constructor, as long as it knows how to provide those types. Hence this calling format can be used for any type where all the ctor dependencies can be resolved by the container.
+``Func<B>`` behaves just like calling ``Resolve<B>()``. That means it's not limited to acting on parameterless constructors on objects - it will wire up constructor parameters, do property injection, and follow the whole lifecycle that ``Resolve<B>()`` does.
 
 Further, lifetime scopes are respected. If you register an object as ``InstancePerDependency()`` and call the ``Func<B>`` multiple times, you'll get a new instance each time; if you register an object as ``SingleInstance()`` and call the ``Func<B>`` to resolve the object more than once, you will get *the same object instance every time*.
 
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -142,7 +142,7 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
 If you would rather control ``B`` disposal yourself all the time, :doc:`register B as ExternallyOwned() <../lifetime/disposal>`.
 
 
-Dynamic Instantiation (`Func<B>`)
+Dynamic Instantiation (Func<B>)
 -------------------------------
 Using an *auto-generated factory* can let you effectively call ``Resolve<B>()`` without tying your component to Autofac. Use this relationship type if you need to create more than one instance of a given service, or if you're not sure if you're going to need a service and want to make the decision at runtime. This relationship is also useful in cases like :doc:`WCF integration <../integration/wcf>` where you need to create a new service proxy after faulting the channel.
 
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -199,7 +199,7 @@ You can also use an *auto-generated factory* to provide parameters when creating
         }
     }
 
-Note that since we're ``.Resolve()``ing the instantiation, rather than actually directly calling the constructor we don't need to declare the parameters in the same order as they appear in the ctor, nor do we necessarily need to provide *all* the parameters listed in that constructor. If some of the constructor's parameters could have been resolved by the Autofac Container already, then those parameters can be omitted from the ``Func`` signature being declared, so you only need to list the types that the Container can't resolve.
+Note that since we're resolving the instance rather than directly calling the constructor we don't need to declare the parameters in the same order as they appear in the constructor definition, nor do we need to provide *all* the parameters listed in the constructor. If some of the constructor's parameters can be resolved by the lifetime scope, then those parameters can be omitted from the ``Func<X, Y, B>`` signature being declared. You only *need* to list the types that the scope can't resolve.
 
 Alternatively, you can use this approach to override a constructor parameter that *would* otherwise have been resolved from the container, with a concrete instance already in hand.
 
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -144,7 +144,13 @@ If you would rather control ``B`` disposal yourself all the time, :doc:`register
 
 Dynamic Instantiation (Func<B>)
 -------------------------------
-Using an *auto-generated factory* can let you effectively call ``Resolve<B>()`` without tying your component to Autofac. Use this relationship type if you need to create more than one instance of a given service, or if you're not sure if you're going to need a service and want to make the decision at runtime. This relationship is also useful in cases like :doc:`WCF integration <../integration/wcf>` where you need to create a new service proxy after faulting the channel.
+Using an *auto-generated factory* can let you resolve a new `B` programmatically within the control flow of your program, without requiring a direct dependency on the Autofac library. Use this relationship type if:
+
+* You need to create more than one instance of a given service.
+* You want to specifically control when the setup of the service occurs.
+* You're not sure if you're going to need a service and want to make the decision at runtime.
+
+This relationship is also useful in cases like :doc:`WCF integration <../integration/wcf>` where you need to create a new service proxy after faulting the channel.
 
 Note that whilst this could be used to simply invoke a parameterless constructor of the target type ``B``, that isn't all it's limited to. As stated above, invoking the ``Func<B>`` is more akin to calling ``Resolve<B>``, and hence Autofac can wire up other types passed into the constructor, as long as it knows how to provide those types. Hence this calling format can be used for any type where all the ctor dependencies can be resolved by the container.
 
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -237,7 +237,7 @@ Internally, Autofac determines what values to use for the constructor args solel
 
 **Lifetime scopes are respected** using this relationship type, just as they are when using :doc:`delegate factories <../advanced/delegate-factories>`. If you register an object as ``InstancePerDependency()`` and call the ``Func<X, Y, B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<X, Y, B>`` to resolve the object more than once, you will get *the same object instance every time regardless of the different parameters you pass in.* Just passing different parameters will not break the respect for the lifetime scope.
 
-As noted above, a consequence of Parameterised Instantiation treating its args as TypedParameters is that **you cannot effectively use auto-generated function factories to invoke constructors which have duplicate types in their parameter list.** For example, say you have a type like this:
+As noted above, ``Func<X, Y, B>`` treats arguments as ``TypedParameter`` so you can't have duplicate types in the parameter list. For example, say you have a type like this:
 
 .. sourcecode:: csharp
 
--- a/docs/resolve/relationships.rst
+++ b/docs/resolve/relationships.rst
@@ -227,7 +227,7 @@ Example:
         }
     }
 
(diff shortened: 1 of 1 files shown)


docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
 If you would rather control ``B`` disposal yourself all the time, :doc:`register B as ExternallyOwned() <../lifetime/disposal>`.
 
 
-Dynamic Instantiation (Func<B>)
+Dynamic Instantiation (`Func<B>`)
```
tillig: RST uses double backticks for code, but they don't work well in headers.
```suggestion
Dynamic Instantiation (Func<B>)
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
+Note that whilst this could be used to simply invoke a parameterless constructor of the target type ``B``, that isn't all it's limited to. As stated above, invoking the ``Func<B>`` is more akin to calling ``Resolve<B>``, and hence Autofac can wire up other types passed into the constructor, as long as it knows how to provide those types. Hence this calling format can be used for any type where all the ctor dependencies can be resolved by the container.
+
+Another corollary of this behaving like calling ``Resolve<B>()`` is that **Lifetime scopes are respected** using this relationship type. If you register an object as ``InstancePerDependency()`` and call the ``Func<B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<B>`` to resolve the object more than once, you will get *the same object instance every time*.
```
tillig: I see the clarification here is that `Func<B>` is not just for parameterless constructors - it's identical in almost every respect to calling `Resolve<B>`. I'm curious if we can be just slightly more succinct in explaining that.

[It doesn't seem suggestions work on multiple lines, at least in the preview. This is a suggestion for the 149-151 lines.]

Maybe something like...

```suggestion
``Func<B>`` behaves just like calling ``Resolve<B>()``. That means it's not limited to acting on parameterless constructors on objects - it will wire up constructor parameters, do property injection, and follow the whole lifecycle that ``Resolve<B>()`` does.

Further, lifetime scopes are respected. If you register an object as ``InstancePerDependency()`` and call the ``Func<B>`` multiple times, you'll get a new instance each time; if you register an object as ``SingleInstance()`` and call the ``Func<B>`` to resolve the object more than once, you will get *the same object instance every time*.
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
+          var b = _newB();
           b.DoSomething();
       }
     }
 
 
 Parameterized Instantiation (Func<X, Y, B>)
 -------------------------------------------
-You can also use an *auto-generated factory* to pass strongly-typed parameters to the resolution function. This is an alternative to :doc:`passing parameters during registration <../register/parameters>` or :doc:`passing during manual resolution <../resolve/parameters>`:
+You can also use an *auto-generated factory* to provide parameters when creating an new instance of the object, where the constructor of the object calls for some additional parameters. As in the case of Dynamic Instantiation above, we're still doing something equivalent to calling ``.Resolve<B>()``, but in this case we're passing strongly-typed parameters to the resolution function. This is an alternative to :doc:`passing parameters during registration <../register/parameters>` or :doc:`passing during manual resolution <../resolve/parameters>`:
```
tillig: Maybe:

```suggestion
You can also use an *auto-generated factory* to provide parameters when creating an new instance of the object, where the constructor of the object calls for some additional parameters. While the ``Func<B>`` relationship is similar to ``Resolve<B>()``, the ``Func<X, Y, B>`` relationship is like calling ``Resolve<B>(TypedParameter.From<X>(x), TypedParameter.From<Y>(y))`` - a resolve operation that has typed parameters. This is an alternative to :doc:`passing parameters during registration <../register/parameters>` or :doc:`passing during manual resolution <../resolve/parameters>`:
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
+
+    public class A
+    {
+        Func<int, P, B> _newB;
+
+        public A(Func<int, P, B> bFactory) { _newB = bFactory }
+
+        public void M(P existingPea)
+        {
+            var b = _newB(42, existingPea); //the needed Q and R will be resolved by Autofac, but not the P since the existingPea will be used.
```
tillig: Lines get a little long with inline comments.

```suggestion
            // The Q and R will be resolved by Autofac, but P will be existingPea instead.
            var b = _newB(42, existingPea);
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
             b.DoSomething();
         }
     }
 
-Internally, Autofac treats these as typed parameters. What that means is that **auto-generated function factories cannot have duplicate types in the input parameter list.** For example, say you have a type like this:
+Internally, Autofac determines what values to use for the constructor args, solely based on the types, and behaves as though we've temporarily defined the input values as the objects that should be used to resolve those types. A direct consequence of this is that  **auto-generated function factories cannot have duplicate types in the input parameter list.** See below for further notes on this.
```
tillig:
```suggestion
Internally, Autofac determines what values to use for the constructor args solely based on the type and behaves as though we've temporarily defined the input values for resolution. A consequence of this is that  **auto-generated function factories cannot have duplicate types in the input parameter list.** See below for further notes on this.
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
+
+
+**Lifetime scopes are respected** using this relationship type, just as they are when using delegate factories. If you register an object as ``InstancePerDependency()`` and call the ``Func<X, Y, B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<X, Y, B>`` to resolve the object more than once, you will get *the same object instance every time regardless of the different parameters you pass in.* Just passing different parameters will not break the respect for the lifetime scope.
+
+Parameterized Instantiation on types with duplicate Types in their constructor args.
+-------------------------------------------
+As noted above, a consequence of Parameterised Instantiation treating its args as TypedParameters is that **you cannot effectively use auto-generated function factories to invoke constructors which have duplicate types in their parameter list.** For example, say you have a type like this:
```
tillig: It feels redundant to have this here when it's basically a copy/paste of text we just read two paragraphs ago.

```suggestion
As noted above, ``Func<X, Y, B>`` treats arguments as ``TypedParameter`` so you can't have duplicate types in the parameter list. For example, say you have a type like this:
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
     }
 
-Internally, Autofac treats these as typed parameters. What that means is that **auto-generated function factories cannot have duplicate types in the input parameter list.** For example, say you have a type like this:
+Internally, Autofac determines what values to use for the constructor args, solely based on the types, and behaves as though we've temporarily defined the input values as the objects that should be used to resolve those types. A direct consequence of this is that  **auto-generated function factories cannot have duplicate types in the input parameter list.** See below for further notes on this.
+
+
+**Lifetime scopes are respected** using this relationship type, just as they are when using delegate factories. If you register an object as ``InstancePerDependency()`` and call the ``Func<X, Y, B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<X, Y, B>`` to resolve the object more than once, you will get *the same object instance every time regardless of the different parameters you pass in.* Just passing different parameters will not break the respect for the lifetime scope.
```
tillig: By moving this further away from the discussion about delegate factories, the "just as they are..." part gets disconnected. We need to link to add that context back.

```suggestion
**Lifetime scopes are respected** using this relationship type, just as they are when using :doc:`delegate factories <../advanced/delegate-factories>`. If you register an object as ``InstancePerDependency()`` and call the ``Func<X, Y, B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<X, Y, B>`` to resolve the object more than once, you will get *the same object instance every time regardless of the different parameters you pass in.* Just passing different parameters will not break the respect for the lifetime scope.
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
 
         public void M()
         {
-            var b = _b(42, "http://hel.owr.ld");
+            var b = _newB(42, "http://hell.owor.ld");
+            b.DoSomething();
+        }
+    }
+
+Note that since we're ``.Resolve()``ing the instantiation, rather than actually directly calling the constructor we don't need to declare the parameters in the same order as they appear in the ctor, nor do we necessarily need to provide *all* the parameters listed in that constructor. If some of the constructor's parameters could have been resolved by the Autofac Container already, then those parameters can be omitted from the ``Func`` signature being declared, so you only need to list the types that the Container can't resolve.
```
tillig: Inline code attached to other things doesn't really render well. Trying to unify references to the function type so it's consistent across all of them (`Func<B>` or `Func<X, Y, B>` rather than also having `Func`). Using "lifetime scope" instead of "container" to drive home that lifetime scope is obeyed. (Registrations can be added to the lifetime scope that don't exist in the container.)

```suggestion
Note that since we're resolving the instance rather than directly calling the constructor we don't need to declare the parameters in the same order as they appear in the constructor definition, nor do we need to provide *all* the parameters listed in the constructor. If some of the constructor's parameters can be resolved by the lifetime scope those parameters can be omitted from the ``Func<X, Y, B>`` signature being declared. You only need to list the types that the scope can't resolve.
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
+Internally, Autofac determines what values to use for the constructor args, solely based on the types, and behaves as though we've temporarily defined the input values as the objects that should be used to resolve those types. A direct consequence of this is that  **auto-generated function factories cannot have duplicate types in the input parameter list.** See below for further notes on this.
+
+
+**Lifetime scopes are respected** using this relationship type, just as they are when using delegate factories. If you register an object as ``InstancePerDependency()`` and call the ``Func<X, Y, B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<X, Y, B>`` to resolve the object more than once, you will get *the same object instance every time regardless of the different parameters you pass in.* Just passing different parameters will not break the respect for the lifetime scope.
+
+Parameterized Instantiation on types with duplicate Types in their constructor args.
+-------------------------------------------
```
tillig: Remove this header - it shouldn't be a whole new section that shows up in the table of contents. This is another where it appears a suggestion over multiple lines doesn't work.

```suggestion
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
+        public void M(P existingPea)
+        {
+            var b = _newB(42, existingPea); //the needed Q and R will be resolved by Autofac, but not the P since the existingPea will be used.
             b.DoSomething();
         }
     }
 
-Internally, Autofac treats these as typed parameters. What that means is that **auto-generated function factories cannot have duplicate types in the input parameter list.** For example, say you have a type like this:
+Internally, Autofac determines what values to use for the constructor args, solely based on the types, and behaves as though we've temporarily defined the input values as the objects that should be used to resolve those types. A direct consequence of this is that  **auto-generated function factories cannot have duplicate types in the input parameter list.** See below for further notes on this.
+
```
tillig, line 237: Remove extra whitespace line.

```suggestion
```

tillig (changes requested):
I think on the whole this is a good set of updates to help clarify what the two `Func`-based relationships do.

I've made some suggestions to:
- Fix some RST vs. Markdown things. RST is kind of a pain.
- Tighten up the verbiage a little. The docs are already long enough that folks don't read them and ask questions about things clearly answered in the docs; keeping it tight means less to read and more likelihood folks will RTFM. 😆 
- Tie things together where content moved around and lost context.

docs/resolve/relationships.rst
Brondahl: Happy with the change around the "a consequence of ... is" clause.

But I was keen on the 2nd alternative phrasing of the bold section, as it's tying it back to the actual constructors.

It's making the point that "it's not really a problem with the Func<...> signature; it's a problem with the constructor signature. If you've got duplicate types in the constructor, then that's just not very compatible with this usage."

Thoughts?
I've made my case, so if it doesn't sway you then I'm happy to just apply this Suggestion as-is :)

docs/resolve/relationships.rst
Brondahl: Just to check, supposing we overode a type that could otherwise be resolved, but that type was also being resolved elsewhere in the dependency chain, as a child dependency of some other auto-resolved ctor parameter, what would happen?

i.e. imagine:

```
    class A
    {
        public A(B b, C c)
        {
            CapturedBee = b;
            CapturedSea = c;
        }
        public B CapturedBee;
        public C CapturedSea;
    }

    class B
    {
        public B(C c)
        {
            CapturedSubSea= c;
        }
        public C CapturedSubSea;
    }

    class C
    {
        public C() {}
    }

    class Elsewhere
    {
        private Func<C, A> _newA;
        public A(Func<C, A> aFactory) { _newA = aFactory; }

        public void DoStuff(C existingSea)
        {
            var a = _newA(existingSea);

            var directResolution //true.
                = Object.ReferenceEquals(existingSea, a.CapturedSea);

            var indirectResolution // ???
                = Object.ReferenceEquals(existingSea, a.CapturedBee.CapturedSubSea);
        }
    }
```
What will `indirectResolution` be?

docs/resolve/relationships.rst
Brondahl: Note `If [...] parameters can be resolved by the lifetime scope those parameters can be omitted [...]` still benefits from the `, then` in the middle to make the sentence easier to parse.

```suggestion
Note that since we're resolving the instance rather than directly calling the constructor we don't need to declare the parameters in the same order as they appear in the constructor definition, nor do we need to provide *all* the parameters listed in the constructor. If some of the constructor's parameters can be resolved by the lifetime scope, then those parameters can be omitted from the ``Func<X, Y, B>`` signature being declared. You only *need* to list the types that the scope can't resolve.
```

Happy?

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
+
+**Lifetime scopes are respected** using this relationship type, just as they are when using delegate factories. If you register an object as ``InstancePerDependency()`` and call the ``Func<X, Y, B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<X, Y, B>`` to resolve the object more than once, you will get *the same object instance every time regardless of the different parameters you pass in.* Just passing different parameters will not break the respect for the lifetime scope.
+
+Parameterized Instantiation on types with duplicate Types in their constructor args.
```
Brondahl: Split existing Suggestion to cover all the lines.
```suggestion
```

docs/resolve/relationships.rst
Brondahl: Split existing Suggestion to cover all the lines.
```suggestion
```

docs/resolve/relationships.rst
Brondahl: Splitting existing suggestions across multiple lines
```suggestion
Further, lifetime scopes are respected. If you register an object as ``InstancePerDependency()`` and call the ``Func<B>`` multiple times, you'll get a new instance each time; if you register an object as ``SingleInstance()`` and call the ``Func<B>`` to resolve the object more than once, you will get *the same object instance every time*.
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
...
 
-**Lifetime scopes are respected** using this relationship type. If you register an object as ``InstancePerDependency()`` and call the ``Func<B>`` multiple times, you'll get a new instance each time. However, if you register an object as ``SingleInstance()`` and call the ``Func<B>`` to resolve the object more than once, you will get *the same object instance every time*.
+Note that whilst this could be used to simply invoke a parameterless constructor of the target type ``B``, that isn't all it's limited to. As stated above, invoking the ``Func<B>`` is more akin to calling ``Resolve<B>``, and hence Autofac can wire up other types passed into the constructor, as long as it knows how to provide those types. Hence this calling format can be used for any type where all the ctor dependencies can be resolved by the container.
```
Brondahl: Splitting existing suggestion across multiple lines

```suggestion
``Func<B>`` behaves just like calling ``Resolve<B>()``. That means it's not limited to acting on parameterless constructors on objects - it will wire up constructor parameters, do property injection, and follow the whole lifecycle that ``Resolve<B>()`` does.
```

docs/resolve/relationships.rst
```diff
@@ -142,50 +142,98 @@ This is by design because you wouldn't want one component to dispose the ``B`` o
 If you would rather control ``B`` disposal yourself all the time, :doc:`register B as ExternallyOwned() <../lifetime/disposal>`.
 
 
-Dynamic Instantiation (Func<B>)
+Dynamic Instantiation (`Func<B>`)
 -------------------------------
-Using an *auto-generated factory* can let you effectively call ``Resolve<T>()`` without tying your component to Autofac. Use this relationship type if you need to create more than one instance of a given service, or if you're not sure if you're going to need a service and want to make the decision at runtime. This relationship is also useful in cases like :doc:`WCF integration <../integration/wcf>` where you need to create a new service proxy after faulting the channel.
+Using an *auto-generated factory* can let you effectively call ``Resolve<B>()`` without tying your component to Autofac. Use this relationship type if you need to create more than one instance of a given service, or if you're not sure if you're going to need a service and want to make the decision at runtime. This relationship is also useful in cases like :doc:`WCF integration <../integration/wcf>` where you need to create a new service proxy after faulting the channel.
```
Brondahl: Suggested rephrasing of 'preceding sentence'. See conversation above.

```suggestion
Using an *auto-generated factory* can let you resolve a new `B` programmatically within the control flow of your program, without requiring a direct dependency on the Autofac Library. Use this relationship type if:
 * you need to create more than one instance of a given service
 * you want to specifically control when the setup of the service occurs
 * you're not sure if you're going to need a service and want to make the decision at runtime.
This relationship is also useful in cases like :doc:`WCF integration <../integration/wcf>` where you need to create a new service proxy after faulting the channel.
```

docs/resolve/relationships.rst
Brondahl: I like that, but the first sentence of the suggestion is now repeating the start of the preceding sentence. I've suggested a change to that preceding sentence to make it more about intent that implementation, to remove that repetition.

docs/resolve/relationships.rst
Brondahl: FBM. Created amd accepted single-line suggestions to reproduce this. So resolving this comment.

docs/resolve/relationships.rst
Brondahl: Issue with Multi-Line suggestions is irritating, isn't it. :(
I've created single-line suggestions applying your changes, so that we can stick within this review workflow. If you're happy with the adjustments to the preceding sentence, then I'll commit all the changes and Resolve this Conversation.

docs/resolve/relationships.rst
tillig: This looks good. My slightly-nitpicky technical writing classes from college are kicking in, though, so some minor changes to make the bullets consistent and not use "Autofac Library" as a proper noun.

```suggestion
Using an *auto-generated factory* can let you resolve a new `B` programmatically within the control flow of your program, without requiring a direct dependency on the Autofac library. Use this relationship type if:

* You need to create more than one instance of a given service.
* You want to specifically control when the setup of the service occurs.
* You're not sure if you're going to need a service and want to make the decision at runtime.

This relationship is also useful in cases like :doc:`WCF integration <../integration/wcf>` where you need to create a new service proxy after faulting the channel.
```

docs/resolve/relationships.rst
tillig: Looks good. [I'm not sure if I commit the suggestion or you do; there's a button allowing me to do that but I'm a bit scared of where that's going to go. I'll let you do it. ;) ]

docs/resolve/relationships.rst
tillig: Yeah, that's cool.

I filed a complaint about the multiline suggestions thing.

tillig: Yup yup.

docs/resolve/relationships.rst
tillig: Yep. Looks good.

docs/resolve/relationships.rst
tillig: There's a lot of "it depends" because we don't know where the `existingSea` in `DoStuff` came from and we don't know the lifetime scopes in which A, B, and C are registered. If they're all singletons the answer may be different than if they're all instance-per-dependency.

However, parameters are provided to the immediate component being resolved, _not to things down the resolution chain_. [That's actually one of our FAQs.](https://autofac.readthedocs.io/en/latest/faq/injecting-global-parameters.html) For all intents and purposes, the `existingSea` and `capturedSea` are likely not going to be the same. They wouldn't be in any case, whether you were calling `Resolve` directly with parameters or using `Func` relationships.

docs/resolve/relationships.rst
tillig: I'd prefer not to over-bold stuff and, when we do, make the point once in bold and let folks continue reading. Two nearly-identical bold items are going to make the "skimmers" think nothing new was said.

tillig (changes requested):
Looking good. Let me know if there are other questions or updates.

docs/resolve/relationships.rst
Brondahl: Cool. So it's not that you're temporarily overriding the *general* rules for resolving that type. It's specifically overriding *how it should be resolved for the argument of this **particular** ctor*

docs/resolve/relationships.rst
Brondahl: Fair. So remove the bolding of the 2nd version, but keep the words?

tillig: No, I think I prefer my suggestion. Thanks!

Brondahl: no problem. Committed.

Brondahl:
OK, so I think that's all the requested changes, now.

Something about what we've done has confused GH, though? Do you know how to declare that the review is complete and ready to be taken forwards?

tillig:
Basically any time the PR contents change it wants a "new review." No worries, I got it.

tillig (approved):
🎉 🎉 🎉

tillig:
Merged! Thanks for your patience and help on this. It is truly appreciated!

PR #95, opened 2020-01-07
Fix broken link in WCF documentation

klk103:
Looks like this blog post about WCF service initialization has moved since 2006


tillig (approved):
Oh, right, that whole MSDN blog "cleanup" fiasco. Thanks!

PR #96, opened 2020-01-26
Removed empty syntax highlight block from documentation.

Myoxocephalus:
Removed empty syntax highlight block from modules documentation.
Before:
Skärmavbild 2020-01-26 kl  11 49 39
After:
Skärmavbild 2020-01-26 kl  11 49 56


tillig (approved):
Whoops! Thanks!

PR #102, opened 2020-04-01
IWebHostingEnvironment to IWebHostEnvironment

om3n07:
There is no such interface `IWebHostingEnvironment`. The recommended interface in this case should be [IWebHostEnvironment](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.iwebhostenvironment?view=aspnetcore-3.1).

Check out this startup class example from the [Microsoft documentation](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-3.1).


tillig (approved):
Nice catch, thanks!

PR #103, opened 2020-05-04
Clarify use of Populate() with ASP.NET Core 3.0+

RustyF:
I spent a while trying to understand how to use IHttpClientFactory (and Polly) with ASP.NET Core 3.0.  It's now obvious to me that it's not an issue given the new AutofacServiceProvider.  I thought it would be worth clarifying this to save other people's time.

### DIFF
--- a/docs/integration/aspnetcore.rst
+++ b/docs/integration/aspnetcore.rst
@@ -201,6 +201,9 @@ In your Startup class (which is basically the same across all the versions of AS
         app.UseMvc();
       }
     }
+No need to call Populate()
+--------------------------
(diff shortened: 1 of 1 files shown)


alistairjevans:
Thanks @RustyF; clarity is always good.

This feels more like a note rather than a full heading item, would you mind defining as a note instead? Like so:

```rst
.. note::
   Note text
```

tillig:
I think it'd be better if the text was a comment inline in the `ConfigureServices` part of the example where, in ASP.NET 2 it would normally be called. That will catch folks right at the point where they'd be trying to do it, so the help would be contextual.

It should be simple, like:

`// No need to call builder.Populate() - that's done by the AutofacServiceProviderFactory`

Which is actually the second item of feedback - it's the factory calling Populate, not the provider.

Finally, while I understand you had trouble with IHttpClientFactory, I'd prefer to not put a specific class "that requires MS DI to configure" because that's the whole point of the integration. Putting the example there out of context unfortunately raises questions for some types of readers like, "wait, I'm using Serilog, not HTTP client factory, so I do still need to call populate?" (Yeah, I know, but that's Stack Overflow for you.) Or "I didn't put the factory into my app startup and I stopped calling Populate and everything's broken!"

If that's not clear enough, instead of creating a "note" I'd rather see an "upgrading from ASP.NET Core 2 to 3" guide with _all_ the differences as a series of steps (add the provider factory, make sure you're not returning a service provider from ConfigureServices, etc). We get questions about all of those things.

tillig:
I added this as a code comment in the Startup class sample.

PR #105, opened 2020-07-02
Composites documentation

alistairjevans:
Initial go at documentation for the new composites feature.

I've created a new v6 branch in the documentation repo to target for this, we can merge that into master when v6 ships.

### DIFF
--- a/docs/advanced/composites.rst
+++ b/docs/advanced/composites.rst
@@ -1,0 +1,170 @@
+==========
+Composites
+==========
+
+The `composite pattern <https://en.wikipedia.org/wiki/Composite_pattern>`_ allows a collection of objects to be treated in the
+same way as a single instance of that same type of object.
+
+In Autofac 6.0, we added support for expressing the composite pattern for services in your container that have multiple implementations, 
+but should be exposed through a single wrapper class.
+
+You can register a **composite wrapper** registration for a service, that will be returned when a consuming class resolves 
+that service, and can take a collection of the same service as a constructor parameter.
+
+This functionality is particularly useful if you find yourself needing to add an additional provider for a service, but you
+don't want the consuming code of the original provider to a) have to change, or b) know that there are multiple providers.
+
+.. sourcecode:: csharp
+
+    // This is my service with multiple registrations.
+    public interface ILogSink
+    {
+        void WriteLog(string log);
+    }
+
+    // A composite wrapper is just a regular class that injects a collection of the same service.
+    public class CompositeLogSink : ILogSink
+    {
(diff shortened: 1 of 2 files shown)


docs/advanced/composites.rst
```diff
@@ -0,0 +1,170 @@
...
+The `composite pattern <https://en.wikipedia.org/wiki/Composite_pattern>`_ allows a collection of objects to be treated in the
+same way as a single instance of that same type of object.
+
+In Autofac 6.0, we added support for expressing the composite pattern for services in your container that have multiple implementations, 
+but should be exposed through a single wrapper class.
+
+You can register a **composite wrapper** registration for a service, that will be returned when a consuming class resolves 
+that service, and can take a collection of the same service as a constructor parameter.
+
+This functionality is particularly useful if you find yourself needing to add an additional provider for a service, but you
```
tillig:
```suggestion
This functionality is particularly useful if you find yourself needing to add a provider for a service, but you
```

docs/advanced/composites.rst
```diff
@@ -0,0 +1,170 @@
...
+    var container = builder.Build();
+
+    // This will return the composite wrapper and expose the metadata.
+    var logSink = container.Resolve<Meta<ILogSink>>();
+
+Lifetime
+--------
+
+Composite wrappers can have their own lifetime, much like any other registration. However, you should consider the 
+implications of making composite registration's long-living; a `SingleInstance` composite would ignore any additional registrations
```
tillig:
```suggestion
implications of making composite registration's long-living; a ``SingleInstance`` composite would ignore any additional registrations
```

tillig (changes requested):
Couple minor things, otherwise this is great.

I have this feeling in the back of my head that there will be folks who assume Autofac will also do _something_ [unquantified] about error handling with the composites. Like, "I did the foreach over the collection and one of them threw an exception. Dammit, Autofac, Y U NO MAGIC?" But it also seems redundant to have to specify, "Hey, you're writing this composite, it's up to _you_ to control how it works."

docs/advanced/composites.rst
alistairjevans: Also noticed that incorrect apostrophe.

alistairjevans:
Since we're making people write their own wrapper, we're mostly minimising the amount of magic we're actually doing.  It might be different if we were providing an automatic 'this composite wrapper automatically calls all your providers in a foreach'. Then we would need to worry about that error handling.

tillig (approved):
:tada:

PR #106, opened 2020-07-13
Added rebus as integration

dariogriffo:


tillig:
I can see where having a "community extensions" page is probably interesting. I'm not sure a dedicated Rebus page is interesting because it gives the appearance we own it and can support it; but directing folks to other extensions is definitely interesting.

If you restructure this as a "Third-Party Integrations" page, we'd take it.

- Named `third-party-integrations.rst`
- Top-level header "Third-Party Integrations"
- Local contents so folks can see a list of the other integrations on the page.
- Second-level header "Rebus" with a description/link.
- Add a link to the new page to the bottom of the TOC on `integrations/index.rst`

If this isn't interesting to you, I can add it as an issue so we don't forget to create such a page.

dariogriffo:
I will, first time using this doc format, so will take my time, but will push something

tillig:
I'm going to close this due to lack of activity. A new pull request with the changes would still be welcome.

PR #113, opened 2020-08-01
Pooling Extension Docs

### LINKED ISSUE #110: Pooling Documentation
Documentation required on how to use the new Autofac.Pooling library.

alistairjevans:
First pass at the pooling documentation.

Got a section in there to explain clearly what 'capacity' means.  Also added a full implementation of a blocking policy as an example.

### DIFF
--- a/docs/advanced/pooled-instances.rst
+++ b/docs/advanced/pooled-instances.rst
@@ -116,11 +116,126 @@ you can implement a custom :ref:`Pool Policy <pooled-instances-policies>`.
 Pool Capacity
 -------------
 
+Each pooled registration has the notion of a pool capacity.  This capacity defaults to ``Environment.ProcessorCount * 2``,
+but can easily be customised using overloads of the extension methods:
 
+.. sourcecode:: csharp
 
+    // Set a capacity of 100
+    builder.RegisterType<MyCustomConnection>()
+            .As<ICustomConnection>()
+            .PooledInstancePerLifetimeScope(100);
 
+It's important to understand that the capacity of a pool does not place a limit on the number of instances it allocates/activates,
+or can be in use at any one time; instead it limits how many instances are **retained** by the pool.
+
+In practical terms, this means that if your pool capacity is 100, and you currently have 100 instances in use, then
+resolving another instance will just activate a brand new instance of the component, rather than blocking/failing. 
+
+However, if you have 101 instances of the component in use, the next instance that is returned to the pool will be discarded,
+rather than retained. 
+
+In this situation, the ``OnReturnToPool`` method on ``IPooledComponent`` would still be called, but the instance will then immediately be thrown away.
+
+When an instance is discarded by the pool, if the object implements ``IDisposable``, ``Dispose`` will be called.
+
+If you in fact do want your pool to have custom behaviour like blocking until a resource is available, you can implement 
+a custom :ref:`Pool Policy <pooled-instances-policies>`. 
+
+.. note:: 
+
+    The Autofac Pooling behaviour is built on top of the `Object Pool <https://docs.microsoft.com/en-us/aspnet/core/performance/objectpool>`_ implementation
+    available from the Microsoft.Extensions.ObjectPool package.
+
+    The behaviour of that pool informs a lot of the behaviour of Autofac.Pooling.
+
+Matching Lifetime Scopes
+------------------------
+
+In the same way that you can configure a normal registration to be scoped to a :ref:`Matching Lifetime Scope <lifetime-instance-scope-per-matching-lifetime-scope>`,
+you can configure a pooled registration to be scoped in the same way:
+
+.. sourcecode:: csharp
+
+    builder.RegisterType<MyCustomConnection>()
+           .As<ICustomConnection>()
+           .PooledInstancePerMatchingLifetimeScope("tag");
+
+Pooled registrations with a Matching Lifetime Scope result in each tagged scope retrieving its own instance from the pool, and child scopes
+sharing the same pooled instance.
+
+When the tagged Lifetime Scope is disposed, the instance is returned to the pool.
 
 .. _pooled-instances-policies:
 
 Pool Policies
 -------------
+
+If you need some custom behaviour that is invoked when instances are retrieved from, or returned to, the pool, you can implement
+``IPooledRegistrationPolicy<TPooledObject>`` or override ``DefaultPooledRegistrationPolicy<TPooledObject>``.
+
+Here's an example of a simple policy that will block any further requests for pooled instances once the available capacity is used up:
+
+.. sourcecode:: csharp
+
+    public class BlockingPolicy<TPooledObject> : IPooledRegistrationPolicy<TPooledObject>
+        where TPooledObject : class
+    {
+        private readonly SemaphoreSlim _semaphore;
+
+        public BlockingPolicy(int maxConcurrentInstances)
+        {
+            // Create a semaphore with the same 'capacity' as the number of instances.
+            _semaphore = new SemaphoreSlim(maxConcurrentInstances);
+
+            // Specify that the pool should hold the specified concurrent instances.
+            MaximumRetained = maxConcurrentInstances;
+        }
+
+        /// <summary>
+        /// Gets a value indicating the maximum number of items that will be retained in the pool. 
+        /// </summary>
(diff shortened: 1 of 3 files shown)


docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+    public class WorkOperation
+    {
+        // Injects an instance from the pool.
+        public WorkOperation(ICustomConnection customConnection)
+        {
+            // ...
+        }
+    }
+
+When the current :ref:`Lifetime Scope <lifetime-instance-scope-per-lifetime-scope>` ends, the retrieved instance is returned to the pool.
```
tillig:
```suggestion
When the current :ref:`lifetime scope <lifetime-instance-scope-per-lifetime-scope>` ends, the retrieved instance is returned to the pool.
```

tillig: Related: `:ref:` is neat - I didn't know about that. I'll have to use that more.

docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+operation, plus any parameters passed to the resolve.
+
+.. warning:: 
+
+    Any services resolved from the provided ``IComponentContext`` are taken from the **current scope** 
+    accessing the pooled component. This means that any instances you resolve from that ``IComponentContext`` 
+    should be discarded in ``OnReturnToPool`` to prevent memory leaks.
+
+If you cannot modify the component you are pooling, but need to have custom behaviour similar to this,
+you can implement a custom :ref:`Pool Policy <pooled-instances-policies>`.
```
tillig:
```suggestion
you can :ref:`implement a custom pool policy <pooled-instances-policies>`.
```

docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+but can easily be customised using overloads of the extension methods:
+
+.. sourcecode:: csharp
+
+    // Set a capacity of 100
+    builder.RegisterType<MyCustomConnection>()
+            .As<ICustomConnection>()
+            .PooledInstancePerLifetimeScope(100);
+
+It's important to understand that the capacity of a pool does not place a limit on the number of instances it allocates/activates,
```
tillig:
```suggestion
It's important to understand that **the capacity of a pool does not place a limit on the number of instances it allocates/activates**,
```

Maybe a personal thing, this was the huge gotcha for me.

docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+or can be in use at any one time; instead it limits how many instances are **retained** by the pool.
+
+In practical terms, this means that if your pool capacity is 100, and you currently have 100 instances in use, then
+resolving another instance will just activate a brand new instance of the component, rather than blocking/failing. 
+
+However, if you have 101 instances of the component in use, the next instance that is returned to the pool will be discarded,
+rather than retained. 
+
+In this situation, the ``OnReturnToPool`` method on ``IPooledComponent`` would still be called, but the instance will then immediately be thrown away.
```
tillig:
```suggestion
However, if you have 101 instances of the component in use, the next instance that is returned to the pool will be discarded
rather than retained. In this situation, the ``OnReturnToPool`` method on ``IPooledComponent`` would still be called, but the instance will then immediately be thrown away.
```

docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+
+When an instance is discarded by the pool, if the object implements ``IDisposable``, ``Dispose`` will be called.
+
+If you in fact do want your pool to have custom behaviour like blocking until a resource is available, you can implement 
+a custom :ref:`Pool Policy <pooled-instances-policies>`.
```
tillig:
```suggestion
If you, in fact, do want your pool to have custom behaviour like blocking until a resource is available, you can :ref:`implement 
a custom pool policy <pooled-instances-policies>`. 
```

docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+
+When an instance is discarded by the pool, if the object implements ``IDisposable``, ``Dispose`` will be called.
+
+If you in fact do want your pool to have custom behaviour like blocking until a resource is available, you can implement 
+a custom :ref:`Pool Policy <pooled-instances-policies>`. 
+
+.. note:: 
+
+    The Autofac Pooling behaviour is built on top of the `Object Pool <https://docs.microsoft.com/en-us/aspnet/core/performance/objectpool>`_ implementation
+    available from the Microsoft.Extensions.ObjectPool package.
```
tillig:
```suggestion
    available from `the Microsoft.Extensions.ObjectPool package <https://www.nuget.org/packages/Microsoft.Extensions.ObjectPool/>`_.
```

docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+
+    The behaviour of that pool informs a lot of the behaviour of Autofac.Pooling.
+
+Matching Lifetime Scopes
+------------------------
+
+In the same way that you can configure a normal registration to be scoped to a :ref:`Matching Lifetime Scope <lifetime-instance-scope-per-matching-lifetime-scope>`,
```
tillig:
```suggestion
In the same way that you can configure a normal registration to be scoped to a :ref:`matching lifetime scope <lifetime-instance-scope-per-matching-lifetime-scope>`,
```

docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+you can configure a pooled registration to be scoped in the same way:
+
+.. sourcecode:: csharp
+
+    builder.RegisterType<MyCustomConnection>()
+           .As<ICustomConnection>()
+           .PooledInstancePerMatchingLifetimeScope("tag");
+
+Pooled registrations with a Matching Lifetime Scope result in each tagged scope retrieving its own instance from the pool, and child scopes
```
tillig:
```suggestion
Pooled registrations with a matching lifetime scope result in each tagged scope retrieving its own instance from the pool, and child scopes
```

docs/advanced/pooled-instances.rst
```diff
@@ -0,0 +1,241 @@
...
+sharing the same pooled instance.
+
+When the tagged Lifetime Scope is disposed, the instance is returned to the pool.
```
tillig:
```suggestion
When the tagged lifetime scope is disposed, the instance is returned to the pool.
```

tillig (changes requested):
This is good stuff. I have some minor fixes around caps and links, but nothing substantive.

alistairjevans:
Generally speaking, do you not consider things like Lifetime Scope and Pool Policy and similar to be proper nouns when we reference them? I don't mind either way, just checking.

tillig:
I don't because it's not the name of the class implementing it (`LifetimeScope`) it's the name of what it _is_ (lifetime scope). Sort of like how you are Alistair the human. Alistair is your name, human is what you are. I think about it like "I'm going to make a pool policy" vs "I'm going to make a PoolPolicy." I don't know that we have anything other than class names and "Autofac" that I'd capitalize. "I need to register a bunch of components that implement services and have a pool policy."

Looking at Microsoft doc ([arbitrary example](https://docs.microsoft.com/en-us/aspnet/core/tutorials/first-mvc-app/adding-controller?view=aspnetcore-3.1&tabs=visual-studio)), that seems in line. Like, "You need to implement a controller that takes a model and returns a view" rather than "You need to implement a Controller that takes a Model and returns a View." The only time I see the caps on Model, View, and Controller is when it's specifically being set next to the acronym "MVC" to explain it, which makes sense... and even then, later in the docs it's not capped.

I definitely overthink these things. And, yeah, I have technical writing references on the bookshelf behind me. Don't @ me. 😆

tillig:
That's also why I left "Autofac Pooling" on line 19 - it's the name of the package, interchangeable with `Autofac.Pooling`. It's not just _any_ pooling package, it's Autofac Pooling. Though I guess one could argue it could be "Autofac pooling" by virtue of it being a function we're providing rather than the name of the thing itself, like "Autofac property injection."

alistairjevans:
Fair points, I get the rule of not wanting to confuse actual type names.

Nice to have a rule for it.

PR #116, opened 2020-08-12
Update the signature of the example IRegistrationSource

### LINKED ISSUE #112: Updates to Documentation on Registration Sources
Changes to the signature of `IRegistrationSource` requires changes to the relevant documentation pages and code examples.

alistairjevans:
I think this is the only change needed to fix #112.

The existing docs don't go into too much depth on how to use the parameters on `RegistrationsFor`, which I think is basically ok.

### DIFF
--- a/docs/advanced/registration-sources.rst
+++ b/docs/advanced/registration-sources.rst
@@ -125,7 +125,7 @@ Here's the code for the registration source.
     {
       public IEnumerable<IComponentRegistration> RegistrationsFor(
         Service service,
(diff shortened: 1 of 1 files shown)


tillig:
Yup, that's it.

Probably would be good _at some point_ to have a richer example that actually uses the `registrationAccessor` parameter to explain what it's for; or round out why you'd use/not use `IsAdapterForIndividualComponents`; but that's all way outside the scope here.

PR #118, opened 2020-08-20
Initial draft of the pipeline docs

### LINKED ISSUE #107: Pipelines Documentation
Need to write documentation on how pipelines work and how people can define and attach their own middleware.

This will be targeted at advanced users.

alistairjevans:
I think I've covered most of the pertinent things that advanced users may want to know about.

Trying to walk the line between informing as to how things work, without overloading the reader with too much internal detail.


docs/advanced/pipelines.rst
```diff
@@ -0,0 +1,246 @@
...
+
+For advanced customisation scenarios, Autofac allows you to add your own middleware into the pipeline to intercept, short-circuit or extend the existing resolve 
+behaviour.
+
+Service Pipelines vs Registration Pipelines
+-------------------------------------------
+
+An individual resolve request actually ends up invoking two different pipelines. The Service Pipeline, and the Registration Pipeline.  
+
+Each service has its own Service Pipeline, and each registration has its own Registration Pipeline.
```
tillig:
```suggestion
Each :doc:`service <../glossary>` has its own Service Pipeline, and each :doc:`registration <../glossary>` has its own Registration Pipeline.
```

I wonder if the definitions of `service` and `registration` need to be fleshed out a bit on that glossary page to make more sense to a general reader. Like, it's not directly obvious that a `service` is _generally_ an interface, or at least, that's kinda how you can think about it. We talk about some of the terms on the `register/registration.rst` page, too. Maybe the glossary should refer to that page for deeper explanation.

Goal here is that when I read this, "each service has a pipeline, each registration has a pipeline" - what... does "service" and "registration" mean and how are they different?

docs/advanced/pipelines.rst
```diff
@@ -0,0 +1,246 @@
...
+    });
+
+Just like with registration middleware, you can register middleware classes instead of lambdas:
+
+.. code-block:: csharp
+
+    builder.RegisterServiceMiddleware<IMyService>(new MyServiceMiddleware());
+
+Service Middleware Sources
+^^^^^^^^^^^^^^^^^^^^^^^^^^
```
tillig:
```suggestion
--------------------------
```
To avoid the "bullet list of one item" in the contents.

tillig (changes requested):
Minor comments; otherwise I can't think of anything calling out to me that's an obvious miss. This is a great run at this sort of doc and I think we'll uncover more as people get using it and questions come in.

docs/advanced/pipelines.rst
alistairjevans: I've applied the reference to the glossary you suggested here; I've also updated the glossary to give a hint about interfaces or other abstractions for the Service entry in the glossary, and referenced the "Services vs. Components" section in the regstration docs.

tillig (approved):
I think this looks great. Should help folks get a feel for how things have changed.

tillig:
🐻

PR #124, opened 2020-09-29
Update log4net example for compatibility with v6

### LINKED ISSUE #123: log4net.rst is broken
Activated and Preparing dont exist any more

alistairjevans:
I implemented the log4net example 'properly', following v6 patterns.

So we have log4net middleware, and a module that adds middleware to every registration.

Fixes #123

### DIFF
--- a/docs/examples/log4net.rst
+++ b/docs/examples/log4net.rst
@@ -1,57 +1,83 @@
-==========================
-log4net Integration Module
-==========================
+==============================
+log4net Integration Middleware
+==============================
 
-While there is no specific assembly for log4net support, you can easily inject ``log4net.ILog`` values using a very small custom module.
+While there is no specific assembly for log4net support, you can easily inject ``log4net.ILog`` values using a simple piece of middleware, and a very small custom module to
+add that middleware to all your registrations.
 
-This module is also a good example of how to use :doc:`Autofac modules <../configuration/modules>` for more than simple configuration - they're also helpful for doing some more advanced extensions.
+The ``Log4NetMiddleware`` is also a good example of how to use :doc:`pipeline middleware <../advanced/pipelines>`.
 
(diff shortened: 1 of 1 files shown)


tillig (approved):
It didn't occur to me how very different this is. Not sure if the folks following this pattern will immediately understand the breaking change here; I'll add a link to this doc from the breaking changes section of the release notes once it's rendered.

VassilSanych:
MiddlewareModule from the sample is not compatible with builder.RegisterModule<...>

alistairjevans:
How so? I can do the following, and I get all the logging behaviour I expect:

```csharp
builder.RegisterModule(new MiddlewareModule(new Log4NetMiddleware()));

builder.RegisterType<Service1>();

var instance = container.Resolve<Service1>();
```

VassilSanych:
This is not belong dependency type injection. You just create the module in place.

alistairjevans:
I'm sorry, I do not understand; you'll need to give me an example of what you mean.

VassilSanych:
For example, you cannot use this with RegisterAssemblyModules.

VassilSanych:
```
public class Log4NetModule : MiddlewareModule<Log4NetMiddleware> {}

public abstract class MiddlewareModule<T> : Module where T : IResolveMiddleware, new()
{
		private readonly IResolveMiddleware _middleware;
		public MiddlewareModule()
		{
			_middleware = new T();
		}
		protected override void AttachToComponentRegistration(IComponentRegistryBuilder componentRegistryBuilder,
			IComponentRegistration registration)
		{
			// Attach to the registration's pipeline build.
			registration.PipelineBuilding += (sender, pipeline) =>
			{
				// Add our middleware to the pipeline.
				pipeline.Use(_middleware);
			};
		}
	}
```
it will be better as it seems to me

alistairjevans:
For the simple module example in the docs it works though. If you in particular need a variant to make it work with RegisterAssemblyModules, then so be it, but that doesn't belong in the docs.

PR #126, opened 2020-10-20
Better convey separation of concerns with app and dependency startup logic.

tillig:
Based on autofac/Autofac#1215 we wanted to make it more clear that, while we offer the ability to run logic on container build, it'd be best if this mechanism wasn't overused to do things like orchestrate app startup logic or do things that go outside dependency management concerns.


alistairjevans (approved):
🕹

PR #127, opened 2020-10-27
Add example for RegistrationData events migration to 6.x

mjamro:
1. Fix compilation of the example because of:

```
A local or parameter named 'sender' cannot be declared in this scope because that name is used in an enclosing local scope to 
define a local or parameter
```

2. Add example to the migration section. I'm not sure if that's the right replacement for ` args.ComponentRegistration.Activated += (o, c)` however it works correctly.


tillig (approved):
Thanks for the PR!

PR #131, opened 2021-01-26
Fix #128: Clarify delegate factory usage

### LINKED ISSUE #128: Clarify delegate factory docs with more commentary
I've been working on Autofac for years and still get confused on the whole delegate factory thing - how they work, why you'd do that over `Func<X, Y, B>`, and so on. [I recently answered a StackOverflow question](https://stackoverflow.com/questions/64566172/autofac-factories-with-named-parameters/64574886#64574886) with a more detailed explanation of things in the docs like what "matches by name rather than type" really means. I think we should flesh out the delegate factory docs a bit more to make the feature more usable.

tillig:
Update to the existing doc that provides a bit more inline commentary for the code and explains in more explicit terms the example we're looking at and the changes seen along the way.

### DIFF
--- a/docs/advanced/delegate-factories.rst
+++ b/docs/advanced/delegate-factories.rst
@@ -2,22 +2,29 @@
 Delegate Factories
 ==================
 
-Factory adapters provide the instantiation features of the container to managed components without exposing the container itself to them.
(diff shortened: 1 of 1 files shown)


alistairjevans (approved):
LGTM 🐓

PR #132, opened 2021-01-26
Fix #104: Explain lifetime scopes with Startable/AutoActivate

### LINKED ISSUE #104: Update docs for `AutoActivate()`
Because of the problems described in this [issue](https://github.com/autofac/Autofac/issues/1150#issuecomment-644178719), the docs should be updated to make it a bit clearer how and when to use `AutoActivate()`.

tillig:
Expanded on the use of lifetime scopes with `IStartable` and `AutoActivate` to avoid confusion like in autofac/Autofac#1150.

### DIFF
--- a/docs/lifetime/startup.rst
+++ b/docs/lifetime/startup.rst
@@ -289,7 +289,7 @@ Further, **you can't use IStartable or AutoActivate with named lifetime scopes.*
         builder.Build();
     }
 
-If you need to start something in a particular lifetime scope, you need to register it with that lifetime scope at the time of creation.
+If you need to start something in a particular lifetime scope, you need to register it with that lifetime scope at you create the scope (i.e., in the ``BeginLifetimeScope`` call).
 
 .. sourcecode:: csharp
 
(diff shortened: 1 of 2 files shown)


docs/lifetime/startup.rst
```diff
@@ -265,6 +265,56 @@ You'll see from the output that the callbacks and ``OnActivated`` methods execut
...
+        // This WON'T WORK. You'll get a DependencyResolutionException when
+        // the container tries to start the component because the named lifetime
+        // scope doesn't exist.
+        builder.RegisterType<Startable1>()
+               .As<IStartable>()
+               .InstancePerMatchingLifetimeScope("unitofwork");
+        builder.Build();
+    }
+
+If you need to start something in a particular lifetime scope, you need to register it with that lifetime scope at the time of creation.
```
alistairjevans: Perhaps you should say "at the time of creating the scope", for extra clarity?

alistairjevans (dismissed):
Looks good, one minor (and optional) clarity suggestion.

docs/lifetime/startup.rst
tillig: Updated to:

> If you need to start something in a particular lifetime scope, you need to register it with that lifetime scope at you create the scope (i.e., in the ``BeginLifetimeScope`` call).

alistairjevans (approved):
🦜

PR #141, opened 2021-04-28
Update overview docs on lifetime scopes

### LINKED ISSUE #89: Clarify how lifetime scopes relate to dependency resolution
Based on autofac/Autofac#1046 - The documentation explains how to avoid captive dependencies, but there's a concept that hasn't been fully explained, which is how a given component locates dependencies.

For example, if you resolve a singleton, even if it's being resolved from a child lifetime scope, the singleton comes from the root container.

If you have a child lifetime scope where an instance-per-dependency component depends on a singleton, the singleton comes from the root but the instance-per-dependency comes from the child scope.

How do things like instance-per-lifetime-scope interact with various components? If you have a singleton that takes in a `Func<T>` dependency, which lifetime scope is it tied to? We don't really spell these sorts of things out in a clear way, so questions pop up fairly frequently that wonder why singletons aren't somehow getting scoped values injected.

tillig:
Resolves #89 

This updates the main lifetime scope doc to have a deeper dive into how scopes are hierarchical and how dependencies are located. Diagrams are updated and more code snippets are added as analogies to help solidify the concepts.

As part of this I also added some Node.js bits that will auto-generate PlantUML diagrams. This allows us to insert more robust diagrams without contributors requiring any real software install (just `npm install` and `npm watch`); and makes for easier updates of existing diagrams since source is maintained in the fairly common PlantUML format.


tillig:
[Here's a trimmed down zip file with the rendered doc that may be easier to review.](https://github.com/autofac/Documentation/files/6393023/html.zip)

.vscode/extensions.json
```diff
@@ -0,0 +1,9 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
+    "ms-python.python",
+    "lextudio.restructuredtext",
+    "stkb.rewrap",
+    "travisillig.vscode-json-stable-stringify"
```
alistairjevans, line 7: Great flex 😁

docs/lifetime/lifetime-scope-tree.puml
```diff
@@ -0,0 +1,48 @@
...
+' Layout hints:
+' https://crashedmind.github.io/PlantUMLHitchhikersGuide/layout/layout.html
+
+database root [
+  <b>Container / Root Lifetime Scope</b>
+
+  The container is where the primary
+  set of registrations live. It serves as
+  the root lifetime scope and holds
+  references to any singletons so they
```
alistairjevans: I wonder if we should clarify `any singletons` here to be clear it's only singletons _declared_ in this scope. We only just had an issue from someone confused about where singleton instances are retained.

docs/lifetime/index.rst
```diff
@@ -21,35 +62,77 @@ As you work in your application, it's good to remember these concepts so you mak
...
+
+What affects this mechanism is the "lifetime" aspect of "lifetime scope." Some components, like singletons need to be shared across multiple scopes. This affects how dependencies get located. The "basic rules" are:
+
+- A child lifetime scope can get dependencies from parent scopes, but a parent scope may not reach down into a child scope. (You can locate things by moving "up the tree" but you can't move "down the tree.")
+- A component will get its dependencies from the scope that owns the component even if the component is resolved by a scope further down the tree. For example, singletons are owned by the root lifetime scope because they need to be shared across all child scopes. When you resolve a singleton, *all of its dependencies will come from the root scope*. It doesn't matter if you're resolving it from a child scope and have overrides or other things you think the singleton should use. Doing it this way ensures you can't dispose dependencies out from under the singleton; and that you won't create a memory leak by holding references to things in child scopes after the scope is disposed.
```
alistairjevans: As the diagram, I think it would be useful to be clear that singletons are owned by the lifetime scope they are declared in, not just kept in the root container.

docs/lifetime/lifetime-scope-web-example.puml
```diff
@@ -0,0 +1,24 @@
+@startuml
+database "Application Container" as root {
+  file "Logging Service" as logger
+}
+
+database "Web Request 1" as request1 {
```
alistairjevans: Propose calling this `Web Request Scope 1` instead, (same for the other one), just to be absolutely clear this box represents a scope created from the container.

docs/lifetime/index.rst
```diff
@@ -21,35 +62,77 @@ As you work in your application, it's good to remember these concepts so you mak
...
+As web requests come in, lifetime scopes might look like this:
 
-When each request ends, the request lifetime scope ends and the respective order processor gets disposed. The logging service, as a singleton, stays alive for sharing by future requests.
+.. image:: lifetime-scope-web-example.png
 
-You can dive deeper on lifetime scopes in `Nick Blumhardt's Autofac lifetime primer <http://nblumhardt.com/2011/01/an-autofac-lifetime-primer/>`_.
+A rough sequence of events for a web app framework like :doc:`ASP.NET Core <../integration/aspnetcore>` goes like this:
 
+1. When a web request comes in, the web application framework creates a child lifetime scope - the "request lifetime scope."
+2. The web application framework resolves the controller instance from the request lifetime scope. Since the controller is registered as "instance per lifetime scope," it will be a sort of "singleton" within that request but will not be shared with other requests.
```
alistairjevans: It may be a bit risky to mix the use of the word "singleton" in this section, even if it is couched with "sort of". 

(also, comma misplaced)

```suggestion
2. The web application framework resolves the controller instance from the request lifetime scope. Since the controller is registered as "instance per lifetime scope", this instance will be shared by any components within that request but will not be shared with other requests.
```

alistairjevans (changes requested):
Generally great improvements; I've got a couple of suggestions about singleton ownership, and there's a comma that got on the wrong side of a quote.

.vscode/extensions.json
tillig: Plus, the ability to avoid weird JSON diff problems by having that consistent order on elements can't be beat. 😄

docs/lifetime/lifetime-scope-tree.puml
tillig: Corrected this in an updated version.

docs/lifetime/index.rst
tillig: I expanded the singleton discussion into a whole section with code and pictures. That definitely has been a point of confusion, so giving that particular concept some extra attention seemed reasonable.

docs/lifetime/lifetime-scope-web-example.puml
tillig: So let it be written, so shall it be done.

docs/lifetime/index.rst
tillig: Updated phrasing, but I think the comma thing is [US English vs. UK English](https://www.grammarly.com/blog/quotation-marks/) - I'll stand by the placement on that one.

tillig:
I think I got all the suggestions addressed - thanks for the thoughtful review. Let me know if you want me to render a new version for easier reading.

docs/lifetime/index.rst
alistairjevans: US English has commas inside the quote? You guys are bonkers.

tillig: You have _no idea_. Oh, wait, you've probably watched politics for the last few years, you _totally have an idea_. 🤪

docs/lifetime/index.rst
```diff
@@ -87,6 +87,113 @@ Part of the job of the lifetime scope is to :doc:`handle disposal of the compone
...
+    {
+      child1Comp = child1.Resolve<Component>();
+      Console.WriteLine(child1Comp.Name);
+    }
+
+    // You can override the singleton by adding a new
+    // singleton in a child scope. This one will be
+    // owned by the child and will work in its children,
+    // but it doesn't override at the root scope.
+    using (var child2 = container.BeginLifetimeScope(
```
alistairjevans, line 167: Does this actually override the `Component` though? You're only redeclaring the `Dependency` here, but not declaring a new `Component`. Unless I'm missing something obvious.

alistairjevans (commented):
Possible code sample problem, or with my brain?

docs/lifetime/index.rst
alistairjevans: This is the last straw; it was always going to be something to do with grammar.

docs/lifetime/index.rst
tillig: Dammit, I forgot to redeclare. My brain nowadays. Good catch.

tillig: OK, got it.

alistairjevans (approved):
There we go; thanks for adding the extra singleton info, this looks good.

PR #142, opened 2021-06-04
Fix broken links in contributor's guide

leotsarev:


tillig (approved):
Thanks for the catch!

PR #146, opened 2021-11-12
Resolve #135: Clarify which properties are autowired.

### LINKED ISSUE #135: Clarify which properties are auto-wired
## Problem Statement

The `PropertiesAutowired` docs explain that properties get injected but not explicitly _which_ properties (e.g., public, unset).

## Desired Solution

Additional docs that spell out more explicitly which properties get injected.

It might be good to also add info about `InjectUnsetProperties` at the same time. Doesn't appear we have docs on that.

tillig:
Resolves #135. Documentation on `PropertiesAutowired` as well as `InjectUnsetProperties`.


docs/register/prop-method-injection.rst
```diff
@@ -21,7 +21,7 @@ To support :doc:`circular dependencies <../advanced/circular-dependencies>`, use
 
     builder.Register(c => new A()).OnActivated(e => e.Instance.B = e.Context.Resolve<B>());
 
-If the component is a :ref:`reflection component <register-registration-reflection-components>`, use the ``PropertiesAutowired()`` modifier to inject properties:
+If the component is a :ref:`reflection component <register-registration-reflection-components>`, use the ``PropertiesAutowired()`` modifier to inject properties. Using ``PropertiesAutoWired()`` will inject properties that are *writable* and *public*:
```
alistairjevans: Is it worth mentioning the `IPropertySelector` overload, or saying "By default, using `PropertiesAutoWired()` will inject properties that are writeable and public"? Or something?

alistairjevans (dismissed):
Fine to go ahead, just one question about whether to mention the `IPropertySelector`.

docs/register/prop-method-injection.rst
tillig: I added more to the example to show default, delegate, and custom selector uses.

PR #148, opened 2021-11-12
Resolve #115: Add ContravariantRegistrationSource docs.

### LINKED ISSUE #115: Documentation on ContravariantRegistrationSource
Autofac has a hidden gem `ContravariantRegistrationSource` but there is no documentation on it so folks don't know about it. Some docs with an example explaining usage (and why you'd want to use it) would be great.

tillig:
Resolves #115. Adds docs to the registration source page to discuss the contravariant registration source.

I'll admit sometimes I get confused with covariance vs. contravariance, too, so double-check my example. I tried to simplify the [tests for `ContravariantRegistrationSource`](https://github.com/autofac/Autofac/blob/master/test/Autofac.Test/Features/Variance/ContravariantRegistrationSourceTests.cs) down to something digestible but admittedly didn't run the whole thing through a compiler or anything.

### DIFF
--- a/docs/advanced/registration-sources.rst
+++ b/docs/advanced/registration-sources.rst
@@ -31,29 +31,32 @@ The ``ContravariantRegistrationSource`` is helpful in registering types that nee
       void Handle(TCommand command);
     }
 
-    public class CommandAHandler : IHandler<CommandA>
+    public class BaseCommandHandler : IHandler<BaseCommand>
     {
-      public void Handle(CommandA command)
+      public void Handle(BaseCommand command)
       {
+         Console.WriteLine(command.GetType().Name);
       }
     }
 
-    public class CommandA
+    public class BaseCommand
     {
     }
 
-    public class CommandB : CommandA
+    public class DerivedCommand : BaseCommand
     {
     }
 
     var builder = new ContainerBuilder();
     builder.RegisterSource(new ContravariantRegistrationSource());
-    builder.RegisterType<CommandAHandler>().As<IHandler<CommandA>>();
+    builder.RegisterType<BaseCommandHandler>().As<IHandler<BaseCommand>>();
     var container = builder.Build();
 
-    // a and b are both CommandAHandler.
-    var a = container.Resolve<IHandler<CommandA>>();
-    var b = container.Resolve<IHandler<CommandB>>();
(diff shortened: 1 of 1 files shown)


docs/advanced/registration-sources.rst
```diff
@@ -20,6 +20,42 @@ You can use the ``Autofac.Features.ResolveAnything.AnyConcreteTypeNotAlreadyRegi
...
+    }
+
+    var builder = new ContainerBuilder();
+    builder.RegisterSource(new ContravariantRegistrationSource());
+    builder.RegisterType<CommandAHandler>().As<IHandler<CommandA>>();
+    var container = builder.Build();
+
+    // a and b are both CommandAHandler.
+    var a = container.Resolve<IHandler<CommandA>>();
+    var b = container.Resolve<IHandler<CommandB>>();
```
alistairjevans: Might I propose an example slightly closer to contravariance names, and using a similar Handle method to the example in the linked MS doc?
```suggestion
    public class BaseCommandHandler : IHandler<BaseCommand>
    {
      public void Handle(BaseCommand command)
      {
         Console.WriteLine(command.GetType().Name);
      }
    }

    public class BaseCommand
    {
    }

    public class DerivedCommand : BaseCommand
    {
    }

    var builder = new ContainerBuilder();
    builder.RegisterSource(new ContravariantRegistrationSource());
    builder.RegisterType<BaseCommandHandler>().As<IHandler<BaseCommand>>();
    var container = builder.Build();

    // a and b are both BaseCommandHandler.
    var a = container.Resolve<IHandler<BaseCommand>>();
    a.Handle(new BaseCommand()); // prints BaseCommand
    var b = container.Resolve<IHandler<DerivedCommand>>();
    b.Handle(new DerivedCommand()); // prints DerivedCommand
```

alistairjevans (changes requested):
Proposed a change to the example.

docs/advanced/registration-sources.rst
tillig: Sounds good to me. Committed. I wasn't sure how large or small the example should be. Somewhat debated not putting an example and referring people to the unit tests. ;)

alistairjevans: Cruel, but not unusual...

PR #149, opened 2021-11-12
Update to chokidar to remove security issues.

tillig:
The `watch` component being used for PlantUML generation was outdated and relied on insecure components. This updates the solution to `chokidar` which is a fairly de-facto standard for file watching. It should also remove the Dependabot alert about security.


alistairjevans (approved):
👍

PR #156, opened 2022-04-07
Add documentation for Lambda Generic Registrations, and fix async disposal docs issue

### LINKED ISSUE #154: Is the important part of the IAsyncDisposable section in the disposal.rst still true?
The following part in https://github.com/autofac/Documentation/blob/master/docs/lifetime/disposal.rst seems to be wrong:

If your component only implements IAsyncDisposable, but someone disposes of the scope synchronously, then Autofac will throw an exception, because it does not know how to dispose of your component.

Warnings are shown but no exceptions are thrown:

AUTOFAC: A synchronous Dispose has been attempted, but the tracked object of type 'XXXX' only implements IAsyncDisposable. This will result in an efficient blocking dispose. Consider either implementing IDisposable on 'UserQuery+TestDisposable' or disposing of the scope/container with DisposeAsync.
DisposedAsync

Related code:
https://github.com/autofac/Autofac/blob/be2a437ca048a7534a5ed7936334d1336291aa1d/src/Autofac/Core/Disposer.cs#L51

https://github.com/autofac/Autofac/commit/3b6f6848396c2363dfcc07ece0b28c86a0ec1ba6#diff-24e2ed8f0c27b3ee3221361170a492f1ec2161d73c1dff50596282b064428d69

alistairjevans:
I decided to stick with talking about the original `IComponentContext` variant first, then introducing the new alternative.

Probably best to hold off merging this until we push the next Autofac release.

Also fixed #154

### DIFF
--- a/docs/register/parameters.rst
+++ b/docs/register/parameters.rst
@@ -69,7 +69,14 @@ Parameters with Lambda Expression Components
 
 With lambda expression component registrations, rather than passing the parameter value *at registration time* you enable the ability to pass the value *at service resolution time*. (:doc:`Read more about resolving with parameters. <../resolve/parameters>`)
 
-In the component registration expression, you can make use of the incoming parameters by changing the delegate signature you use for registration. Instead of just taking in an ``IComponentContext`` parameter, take in an ``IComponentContext`` and an ``IEnumerable<Parameter>``:
+In the component registration expression, you can make use of the incoming parameters by using the generic lambda ``Register`` methods:
+
+.. sourcecode:: csharp
+
+    builder.Register((string configSectionName) => new ConfigReader(configSectionName));
+
+If you need access to the full list of parameters, it's available by changing the delegate signature you use for registration.
+Instead of specifying the parameter as an argument to the lambda, take in an ``IComponentContext`` and an ``IEnumerable<Parameter>``:
 
 .. sourcecode:: csharp
 
--- a/docs/register/parameters.rst
+++ b/docs/register/parameters.rst
@@ -73,7 +73,7 @@ In the component registration expression, you can make use of the incoming param
 
 .. sourcecode:: csharp
 
-    builder.Register((string configSectionName) => new ConfigReader(configSectionName));
+    builder.Register((MyConfig config) => new Worker(config));
 
 If you need access to the full list of parameters, it's available by changing the delegate signature you use for registration.
 Instead of specifying the parameter as an argument to the lambda, take in an ``IComponentContext`` and an ``IEnumerable<Parameter>``:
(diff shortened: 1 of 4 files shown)


docs/register/parameters.rst
```diff
@@ -69,7 +69,14 @@ Parameters with Lambda Expression Components
 
 With lambda expression component registrations, rather than passing the parameter value *at registration time* you enable the ability to pass the value *at service resolution time*. (:doc:`Read more about resolving with parameters. <../resolve/parameters>`)
 
-In the component registration expression, you can make use of the incoming parameters by changing the delegate signature you use for registration. Instead of just taking in an ``IComponentContext`` parameter, take in an ``IComponentContext`` and an ``IEnumerable<Parameter>``:
+In the component registration expression, you can make use of the incoming parameters by using the generic lambda ``Register`` methods:
```
tillig, line 72: Should we change the parameter to be something that isn't a `string`? I ask because usually folks don't register a string as a dependency.

```c#
// Generally not a thing
builder.RegisterInstance("configValue");
```

I see that it's a string below in the parameter section, but it may be worth updating both examples to be some sort of configuration object that might be more commonly registered.

docs/register/registration.rst
```diff
@@ -140,6 +140,18 @@ The parameter ``c`` provided to the expression is the *component context* (an ``
 
 Additional dependencies can be satisfied using this context parameter - in the example, ``A`` requires a constructor parameter of type ``B`` that may have additional dependencies.
 
+As well as using ``IComponentContext`` to resolve dependencies in your lambda expression, you can also use the generic ``Register`` overloads to specify your dependencies as a variable number of typed arguments to the lambda, and Autofac will resolve them for you:
+
+.. sourcecode:: csharp
+
+  builder.Register((IDependency1 dep1, IDependency2 dep2) => new Component(dep1, dep2));
```
tillig: I _think_ RST needs four-space indents for source.
```suggestion
    builder.Register((IDependency1 dep1, IDependency2 dep2) => new Component(dep1, dep2));
```

docs/register/registration.rst
```diff
@@ -140,6 +140,18 @@ The parameter ``c`` provided to the expression is the *component context* (an ``
...
+
+You can blend the ``IComponentContext`` and generic approach if you need to make conditional choices, or use methods like ``ResolveNamed``:
+
+.. sourcecode:: csharp
+
+  builder.Register((IComponentContext ctxt, IDependency1 dep1) => new Component(dep1, ctxt.ResolveNamed<IDependency2>("value")));
```
tillig:
```suggestion
    builder.Register((IComponentContext ctxt, IDependency1 dep1) => new Component(dep1, ctxt.ResolveNamed<IDependency2>("value")));
```

docs/register/registration.rst
```diff
@@ -140,6 +140,18 @@ The parameter ``c`` provided to the expression is the *component context* (an ``
 
 Additional dependencies can be satisfied using this context parameter - in the example, ``A`` requires a constructor parameter of type ``B`` that may have additional dependencies.
 
+As well as using ``IComponentContext`` to resolve dependencies in your lambda expression, you can also use the generic ``Register`` overloads to specify your dependencies as a variable number of typed arguments to the lambda, and Autofac will resolve them for you:
+
+.. sourcecode:: csharp
+
+  builder.Register((IDependency1 dep1, IDependency2 dep2) => new Component(dep1, dep2));
+
+You can blend the ``IComponentContext`` and generic approach if you need to make conditional choices, or use methods like ``ResolveNamed``:
```
tillig:
```suggestion
You can blend the ``IComponentContext`` and generic approach if you need to make conditional choices, or use methods like ``ResolveNamed``. Just add the ``IComponentContext`` as the first parameter to the lambda:
```

Figured it might be worth clarifying that folks probably shouldn't just jam it at the end or whatever. With 20 overloads (!) now for the function, it might not be obvious.

tillig (changes requested):
Couple clarifying questions. Agreed, probably shouldn't merge this until we release the updated package.

docs/register/parameters.rst
alistairjevans: I've given a more "config object" style example.

tillig (approved):
🧑‍🍳

PR #157, opened 2022-06-19
Update parameters.rst

snowbear2014:
To my understanding, The Worker constructor need a MyConfig instance, and we're trying to resolver a Worker instead of a MyConfig in the end.

### DIFF
--- a/docs/register/parameters.rst
+++ b/docs/register/parameters.rst
@@ -95,4 +95,4 @@ When :doc:`resolving with parameters <../resolve/parameters>`, your lambda will 
       SomeValue = "../"
     };
 
-    var worker = scope.Resolve<Worker>(new NamedParameter(config, customConfig));
+    var worker = scope.Resolve<Worker>(new NamedParameter("config", customConfig));
--- a/docs/register/parameters.rst
+++ b/docs/register/parameters.rst
@@ -95,4 +95,4 @@ When :doc:`resolving with parameters <../resolve/parameters>`, your lambda will 
       SomeValue = "../"
     };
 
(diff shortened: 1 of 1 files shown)


docs/register/parameters.rst
```diff
@@ -95,4 +95,4 @@ When :doc:`resolving with parameters <../resolve/parameters>`, your lambda will
       SomeValue = "../"
     };
 
-    var reader = scope.Resolve<MyConfig>(new NamedParameter(config, customConfig));
\ No newline at end of file
+    var worker = scope.Resolve<Worker>(new NamedParameter(config, customConfig));
```
tillig:
```suggestion
    var worker = scope.Resolve<Worker>(new NamedParameter("config", customConfig));
```

tillig: `"config"` on line 87 also should be in quotes.

tillig (changes requested):
Nice find! If you wouldn't mind, it looks like there are a couple of minor related issues: for named parameters, the name of the parameter is a string `"in quotes"` and it looks like that got missed, too.

docs/register/parameters.rst
snowbear2014: good catch, changed.

tillig: ...also on line 87. I can't make a comment on that line or I'd do it. Thanks!

snowbear2014: sorry I missed the one on line 87. done, thanks.

tillig (approved):
Super! Thanks!

PR #162, opened 2022-12-07
Created troubleshooting section with exception docs

tillig:
This PR is in preparation for a PR in core Autofac where we add links in exception messages back to specific docs.

In order to do that, I needed to break the single-page troubleshooting section out into its own section/folder and then under that have a folder of exception docs. Some things got shuffled around/renamed, but generally that's the big change.

I did fix a couple of things to address weird spelling issues (e.g., `gsnuget.png` => `gs-nuget.png`) while I was in here.

I don't _think_ this will break anything - all the links internally check out during the build. External links to the troubleshooting section may break, unclear if that matters.

I only have three exception docs, but I think they're the primary ones. We can always follow this pattern to add more; it's been a long time coming and I'd rather get something out the door than wait for it to be perfect.


docs/troubleshooting/exceptions/no-found-constructors-can-be-invoked.rst
```diff
@@ -0,0 +1,34 @@
...
+===============
+
+::
+
+    None of the constructors found on type 'MyApp.MyComponent' can be invoked
+    with the available services and parameters:
+    Cannot resolve parameter 'MyApp.IMyDependency myDep' of constructor
+    'Void .ctor(MyApp.IMyDependency)'.
+
+This error means that Autofac tried to instantiate a component using its constructor, one or more of the dependencies listed as parameters in the constructor cannot be resolved from the container. The message tells you what parameters could not be resolved.
```
alistairjevans:
```suggestion
This error means that Autofac tried to instantiate a component using its constructor, but one or more of the dependencies listed as parameters in the constructor cannot be resolved from the container. The message tells you what parameters could not be resolved.
```

docs/troubleshooting/exceptions/no-scope-with-matching-tag.rst
```diff
@@ -0,0 +1,27 @@
...
+- **If you've created custom per-request semantics** in your own application, perhaps creating something similar to ``InstancePerRequest``, likely there is some code registering something as ``InstancePerRequest`` but the service is being resolved outside a request lifetime scope. This can happen if you try to share a registration module between something that has per-request lifetime scopes and something that doesn't, like a web app and a background processing job; or if code runs at application startup before there are requests and tries to use per-request objects.
+- **If you register some components with a matching lifetime scope and some without** it can be problems. For example, say you register one ``IMyService`` as a singleton and then register a second ``IMyService`` as ``InstancePerMatchingLifetimeScope``. If you try to resolve ``IEnumerable<IMyService>`` outside a matching lifetime scope, you'll get an error. Even though one of the ``IMyService`` could be resolved, *not all* of them can. ``InstancePerMatchingLifetimeScope`` is not a *filter* to allow an instance to *sometimes* be resolved.
```
alistairjevans:
```suggestion
- **If you register some components with a matching lifetime scope and some without** it can cause problems. For example, say you register one ``IMyService`` as a singleton and then register a second ``IMyService`` as ``InstancePerMatchingLifetimeScope``. If you try to resolve ``IEnumerable<IMyService>`` outside a matching lifetime scope, you'll get an error. Even though one of the ``IMyService`` could be resolved, *not all* of them can. ``InstancePerMatchingLifetimeScope`` is not a *filter* to allow an instance to *sometimes* be resolved.
```

docs/troubleshooting/index.rst
```diff
@@ -0,0 +1,42 @@
...
+=============================
+
+If you're really having trouble and you haven't been able to get `an answer on StackOverflow <https://stackoverflow.com/questions/tagged/autofac>`_ you may want to try doing a little more in-depth debugging/troubleshooting on your own. Here are some tips on doing that.
+
+Exceptions
+----------
+
+The exceptions generated by Autofac really try to point you in the right direction with respect to what could be going wrong.
+
+**Don't panic! Stop and read what it's actually telling you.** Here are some tips for reading Autofac exceptions that can save you a lot of time:
```
alistairjevans, line 12: ezgif com-gif-maker

alistairjevans (changes requested):
Two small english changes and an obligatory hitch-hikers reference.

tillig:
Committed changes as suggested. (Yeah, I snuck a little Hitchhiker's Guide in there. I wish I could have put it in larger, friendlier letters.)

alistairjevans (approved):
📚

PR #163, opened 2023-01-10
Docs for constructor finder not finding ctors.

tillig:
Now that the reporting for "no constructors found" is centralized, I'll add this document here, then add the doc link to the `IConstructorFinder` message once I have it set up.


alistairjevans (approved):
LGTM

PR #164, opened 2023-03-05
Release 7.0

alistairjevans:
Docs for:

- Required Properties
- Assembly Load Context support
- Obsolete RegisterGeneratedFactory

### DIFF
--- a/docs/advanced/scopes-loadcontexts.rst
+++ b/docs/advanced/scopes-loadcontexts.rst
@@ -1,0 +1,71 @@
+AssemblyLoadContext and Lifetime Scopes
+=======================================
+
+In .NET Core and .NET 5+, the `the AssemblyLoadContext was introduced <https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext>`_.
+
+This allowed developers to dynamically load and unload assemblies from their application, and is very useful for developers writing applications with plugin-based architectures.
+
+In order to unload the assemblies in an ``AssemblyLoadContext`` when a plugin is no longer needed, and reclaim the memory required by that plugin, no references to types in that assembly can be held by anything outside the loaded context. This includes Autofac, which by default holds a variety of internal references and caches for types that have been registered.
+
+As of Autofac 7.0, we've added support for indicating to Autofac that a given lifetime scope represents types loaded for a given ``AssemblyLoadContext``; when a lifetime scope created for a specific ``AssemblyLoadContext`` is unloaded, Autofac will make a *best-effort* attempt to remove all references we hold for types from the loaded context, to allow the ``AssemblyLoadContext`` to be unloaded.
+
+You indicate that a lifetime scope is for an ``AssemblyLoadContext`` with the new ``BeginLoadContextLifetimeScope`` method. Here's a full example:
+
+.. sourcecode:: csharp
+
+    //
+    // In PluginDefinition project.
+    //
+
+    public interface IPlugin
+    {
+      void DoSomething();
+    }
+
+    //
+    // In MyPlugin project.
+    //
+
+    public class MyPlugin : IPlugin
+    {
+        public void DoSomething()
+        {
+          Console.WriteLine("Hello World");
+        }
+    }
+
+    //
+    // In the application.
+    //
+
+    var builder = new ContainerBuilder();
+
+    // Components defined in the "Default" load context will be available in load context lifetime scopes.
+    builder.RegisterType<DefaultComponent>();
+
+    var container = builder.Build();
+
+    var loadContext = new AssemblyLoadContext("PluginContext", isCollectible: true);
+
+    using (var scope = container.BeginLoadContextLifetimeScope(loadContext, builder =>
+    {
+      var pluginAssembly = loadContext.LoadFromAssemblyPath("plugins/MyPlugin.dll");
+
+      builder.RegisterAssemblyTypes(pluginAssembly);
+    }))
+    {
+      // Because of default load context behaviour, because we have already loaded
+      // PluginDefinition into the application to use IPlugin, the loaded MyPlugin assembly
+      // will share the same type.
+      var plugin = scope.Resolve<IPlugin>();
+
+      plugin.DoSomething();
+    }
+
+    loadContext.Unload();
+
+.. note::
+
+  If you capture a reference to any resolved components, or any types in the loaded assembly, outside Autofac it's highly likely you won't be able to unload your load context.
+
+  AssemblyLoadContexts are tricky to use in such a way that unloading is guaranteed every time (whether using Autofac or not). See the dotnet documentation on `troubleshooting unloadability <https://learn.microsoft.com/en-us/dotnet/standard/assembly/unloadability#troubleshoot-unloadability-issues>`_ if you run into problems.
--- a/docs/advanced/scopes-loadcontexts.rst
+++ b/docs/advanced/scopes-loadcontexts.rst
@@ -55,7 +55,7 @@ You indicate that a lifetime scope is for an ``AssemblyLoadContext`` with the ne
     }))
     {
       // Because of default load context behaviour, because we have already loaded
-      // PluginDefinition into the application to use IPlugin, the loaded MyPlugin assembly
+      // PluginDefinition into the *application* so we can reference IPlugin, the loaded MyPlugin assembly
       // will share the same type.
       var plugin = scope.Resolve<IPlugin>();
 
@@ -69,3 +69,7 @@ You indicate that a lifetime scope is for an ``AssemblyLoadContext`` with the ne
   If you capture a reference to any resolved components, or any types in the loaded assembly, outside Autofac it's highly likely you won't be able to unload your load context.
 
   AssemblyLoadContexts are tricky to use in such a way that unloading is guaranteed every time (whether using Autofac or not). See the dotnet documentation on `troubleshooting unloadability <https://learn.microsoft.com/en-us/dotnet/standard/assembly/unloadability#troubleshoot-unloadability-issues>`_ if you run into problems.
+
+You can create additional lifetime scopes from your "load context scope" using the regular `BeginLifetimeScope` method, without needing to further track your load context.
+
+That means you can load a plugin, then the plugin can resolve `ILifetimeScope` and create new scopes, with all the assembly metadata being isolated to that initial "load context scope".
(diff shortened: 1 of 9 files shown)


docs/advanced/scopes-loadcontexts.rst
```diff
@@ -0,0 +1,75 @@
+AssemblyLoadContext and Lifetime Scopes
+=======================================
+
+In .NET Core and .NET 5+, the `the AssemblyLoadContext was introduced <https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext>`_.
+
+This allowed developers to dynamically load and unload assemblies from their application, and is very useful for developers writing applications with plugin-based architectures.
```
tillig:
```suggestion
In .NET Core `the AssemblyLoadContext was introduced <https://learn.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext>`_, allowing developers to dynamically load and unload assemblies from their application. This is very useful for developers writing applications with plugin-based architectures.
```

I don't think there's a .NET 5+ that isn't effectively .NET Core.

No action right now, but [this kinda ties into the issue we have where we need more docs about plugins and assembly loading](https://github.com/autofac/Documentation/issues/101). Again, nothing to do right now, just "filing for future reference."

docs/advanced/scopes-loadcontexts.rst
```diff
@@ -0,0 +1,75 @@
...
+
+In order to unload the assemblies in an ``AssemblyLoadContext`` when a plugin is no longer needed, and reclaim the memory required by that plugin, no references to types in that assembly can be held by anything outside the loaded context. This includes Autofac, which by default holds a variety of internal references and caches for types that have been registered.
```
tillig:
```suggestion
In order to unload the assemblies in an ``AssemblyLoadContext`` when an assembly is no longer needed, no references to types in that assembly can be held by anything outside the loaded context. This includes Autofac, which by default holds a variety of internal references and caches for types that have been registered.
```

Didn't want to use "plugin" and "assembly" interchangeably here.

docs/advanced/scopes-loadcontexts.rst
```diff
@@ -0,0 +1,75 @@
...
+
+    var container = builder.Build();
+
+    var loadContext = new AssemblyLoadContext("PluginContext", isCollectible: true);
+
+    using (var scope = container.BeginLoadContextLifetimeScope(loadContext, builder =>
+    {
+      var pluginAssembly = loadContext.LoadFromAssemblyPath("plugins/MyPlugin.dll");
+
+      builder.RegisterAssemblyTypes(pluginAssembly);
```
tillig:
```suggestion
      builder.RegisterAssemblyTypes(pluginAssembly).AsImplementedInterfaces();
```

I think you'll need that to resolve it as `IPlugin`.

docs/advanced/scopes-loadcontexts.rst
```diff
@@ -0,0 +1,75 @@
...
+    }))
+    {
+      // Because of default load context behaviour, because we have already loaded
+      // PluginDefinition into the *application* so we can reference IPlugin, the loaded MyPlugin assembly
+      // will share the same type.
```
tillig:
```suggestion
      // The application should reference the PluginDefinition project, which means the
      // default load context will have loaded the IPlugin interface already. When the
      // MyPlugin assembly gets loaded it should share the same type and allow resolution
      // with the common interface.
```

I think this is what you were going for but some of the "because..., because..." got a bit confusing for me.

docs/register/prop-method-injection.rst
```diff
@@ -21,7 +21,118 @@ To support :doc:`circular dependencies <../advanced/circular-dependencies>`, use
...
+.. sourcecode:: csharp
+
+    public class MyComponent
+    {
+      public required ILogger Logger { get; set; }
+
+      public required IConfigReader ConfigReader { get; set; }
+    }
+
+When the component is resolved, Autofac will populate the ``Logger`` and ``ConfigReader`` properties as if they were constructor parameters.
```
tillig:
```suggestion

      public IDatabaseContext Context { get; set; }
    }

When the component is resolved, Autofac will populate the ``Logger`` and ``ConfigReader`` properties as if they were constructor parameters. The ``Context`` property will be treated like a standard property and will not be populated by default.
```

I figured it would be valuable to show that just because some props get required, no, that doesn't mean they all suddenly become required. This seems obvious yet is also a question I can easily see someone misunderstanding.

docs/register/prop-method-injection.rst
```diff
@@ -43,12 +154,45 @@ If the component is a :ref:`reflection component <register-registration-reflecti
     // is important!
     builder.RegisterType<C>().PropertiesAutowired(new MyCustomPropSelector());
 
+.. note::
+
+  Properties set on a component because they are ``required`` may lead to duplicate injection of a property if ``PropertiesAutowired`` is also used on that component.  Consider using ``PropertiesAutowired(PropertyWiringOptions.PreserveSetValues)`` to avoid repeating property injection.
```
tillig: I didn't even think about this. Might be worth, at some point, figuring out how to track this so folks don't have to worry about it.

docs/register/prop-method-injection.rst
```diff
@@ -43,12 +154,45 @@ If the component is a :ref:`reflection component <register-registration-reflecti
...
+------------------------------
+
 If you have one specific property and value to wire up, you can use the ``WithProperty()`` modifier:
 
 .. sourcecode:: csharp
 
     builder.RegisterType<A>().WithProperty("PropertyName", propertyValue);
 
+Overriding Required Properties
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
tillig:
```suggestion
------------------------------
```

I think this would basically end up creating a "bullet list of one thing" underneath "Manually Specifying Properties." Can't have a bullet list of one.

docs/resolve/relationships.rst
```diff
@@ -394,8 +406,6 @@ Now the function will work:
     var func = scope.Resolve<FactoryDelegate>();
     var obj = func(1, 2, "three");
 
-Another option you have is to use a :doc:`delegate factory, which you can read about in the advanced topics section <../advanced/delegate-factories>`.
-
 Should you decide to use the built-in auto-generated factory behavior (``Func<X, Y, B>``) and only resolve a factory with one of each type, it will work but you'll get the same input for all constructor parameters of the same type.
```
tillig: This extra bit down here is now "under" the `RegisterGeneratedFactory` header but is about `Func<T,U,V>` relationship so it feels kinda weird, like it should pertain to `RegisterGeneratedFactory` but probably belongs up above the `Delegate Factories` header.

docs/whats-new/upgradingfrom6to7.rst
```diff
@@ -0,0 +1,16 @@
+Upgrading from Autofac 6.x to 7.x
+=================================
+
+In the upgrade from Autofac 6.x to 7.x, the amount of breaking change was fairly minimal, and unlikely to impact most users, but there are a couple of distinct changes to make you aware of.
+
+- We have dropped direct support for the `net5.0` TFM (and we recommend you upgrade to `net7.0`), but you can still use this release on .NET 5.0 via our `netstandard2.1` TFM support.
```
tillig:
```suggestion
- We have dropped direct support for the `net5.0` TFM (and we recommend you upgrade to `net6.0` or higher), but you can still use this release on .NET 5.0 via our `netstandard2.1` TFM support.
```

.NET 6 is the LTS right now so it seems safer to recommend "LTS plus".

tillig (changes requested):
A couple suggestions, generally looking good. Thanks for taking the time to do this, I'm excited for the release.

docs/register/prop-method-injection.rst
alistairjevans: Hmm, you're right...might actually make the change pre release to ignore required properties in the classic PropertiesAutowired, might as well fix it now.

tillig: Seems reasonable.

tillig: Looks like this got fixed up in the code.

docs/whats-new/upgradingfrom6to7.rst
```diff
@@ -0,0 +1,16 @@
...
+
+In the upgrade from Autofac 6.x to 7.x, the amount of breaking change was fairly minimal, and unlikely to impact most users, but there are a couple of distinct changes to make you aware of.
+
+- We have dropped direct support for the `net5.0` TFM (and we recommend you upgrade to `net6.0` or higher), but you can still use this release on .NET 5.0 via our `netstandard2.1` TFM support.
+
+- The introduction of :doc:`automatic injection for required properties <../register/prop-method-injection>` means that any existing required properties on components will now be automatically injected.
+
+  If a required property *cannot* be resolved during component activation, an exception will now be thrown.
+
+  Equally, if you previously used ``PropertiesAutowired()`` to populate required properties on components, those properties will now be doubly-resolved, so you may wish to remove ``PropertiesAutowired()`` in favour of the default functionality.
```
tillig: This will need to get updated now that it's not double-populated.

alistairjevans:
Will go through the docs updates now the duplicate stuff is fixed.

alistairjevans:
OK, have updated the PR to reflect the fact that `PropertiesAutowired` no longer duplicates.

Re: the `Func<X, Y, B>` docs, I've moved it, but to be honest I'm not certain what it's trying to say? May be I'm just not sufficiently familiar with parameterised factories; if you have any suggestions @tillig I would appreciate it.

tillig:
Let me try to grab a couple of minutes and commit something to the branch. I need to read it in context and it's going to be a pain to try to figure out via a PR.

tillig (approved):
I updated the registration page and moved the RegisterGeneratedFactory stuff to the Delegate Factories page because... basically it was delegate factories but with an explicit registration line that's no longer needed. Still searchable in the docs, just not on that relationships page.

I consolidated the docs about `Func<X, Y, B>` so maybe it's more concise and clear what it's talking about. It did seem to jump around a lot.

alistairjevans:
Great. Looks like we're ready to go; I'll push the release tomorrow if everything looks shipshape?

tillig:
LET'S DOO ITTTTTTTT :shipit:

tillig:
Oh dammit I hit the button in my excitement. Let me roll it back.

tillig:
OK, I put a PR in that just reverts this commit. Then we can, uh, revert the revert I guess. Sorry. Wasn't thinking.

alistairjevans:
That merge button just sits there, all green and staring at you. I get it. I'll push the release tomorrow then merge the new PR.

PR #166, opened 2023-03-06
Release 7.0

tillig:
Contains the 7.0 release documentation.

I clicked the button too fast on the original 7.0 PR so this is a revert-of-a-revert to get the 7.0 docs in.


tillig:
I did manually roll back the tag v7.0.0 as well and the ReadTheDocs build for 7 was stopped, so it's not showing up in the UI at the moment. I think this should kick off all the right processes.

PR #167, opened 2023-03-07
Update prop-method-injection.rst

TonyValenti:
Suggest required properties use ``public required ... { protected get; init; }``.

### DIFF
--- a/docs/register/prop-method-injection.rst
+++ b/docs/register/prop-method-injection.rst
@@ -32,14 +32,16 @@ For example, given the following type:
 
     public class MyComponent
     {
-      public required ILogger Logger { get; set; }
+      public required ILogger Logger { protected get; init; }
 
-      public required IConfigReader ConfigReader { get; set; }
+      public required IConfigReader ConfigReader { protected get; init; }
 
       public IDatabaseContext Context { get; set; }
     }
 
 When the component is resolved, Autofac will populate the ``Logger`` and ``ConfigReader`` properties as if they were constructor parameters. The ``Context`` property will be treated like a standard property and will not be populated by default.
+
+You can use any valid combination of access modifiers on required properties, however, ``public required ... { protected get; init; }`` is recommended as it provides access and visibility that are similar to constructors: the property is only settable at construction and not publicly visible to other classes.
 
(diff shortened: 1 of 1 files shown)


docs/register/prop-method-injection.rst
```diff
@@ -32,27 +32,29 @@ For example, given the following type:
...
 
-      public required IConfigReader ConfigReader { get; set; }
+      public required IConfigReader ConfigReader { protected get; init; }
 
       public IDatabaseContext Context { get; set; }
     }
 
 When the component is resolved, Autofac will populate the ``Logger`` and ``ConfigReader`` properties as if they were constructor parameters. The ``Context`` property will be treated like a standard property and will not be populated by default.
 
+You can use any valid combination of access modifiers on required properties, however, ``public required ... { protected get; init; }`` is recommended as it provides access and visibility that are similar to constructors: the property is only settable at construction and not publicly visible to other classes.
```
tillig: I'm cool updating the examples, but I'm not sure I want to recommend certain access modifiers. I get the desire, but by recommending it explicitly we end up with questions about why folks _can't_ use something else (they can) or whether they're _supposed to_ treat the properties as different (they don't have to). We generally try to only recommend things when it comes to Autofac usage and patterns.

If we can drop this paragraph, I'll accept the rest of the PR.

tillig (changes requested):
One paragraph to update, otherwise looking good.

docs/register/prop-method-injection.rst
TonyValenti: @tillig Are you OK with changing it to simply:

You can use any valid combination of access modifiers on required properties, however, ``public required ... { protected get; init; }`` is ***used in these examples*** as it provides access and visibility that are similar to constructors: the property is only settable at construction and not publicly visible to other classes.


I want to at least provide the thought process behind why that combination is shown.

TonyValenti: If you're OK with my suggestion, I've updated the PR.  If not, let me know and I'll fully remove that paragraph.

tillig (approved):
I like that new change. Sweet! Thanks!

Issue #169, opened 2023-03-13
Broken links in upgrade documentation


steveoh:
## Describe the Bug

I am upgrading from 6.5.0 to 7.0.0 and have a build time issue with the following code

```c#
builder.Register<ServiceFactory>(ctx => {
  var c = ctx.Resolve<IComponentContext>();
  return t => c.Resolve(t);
});
```

## Steps to Reproduce

I went to the [upgrade doc](https://docs.autofac.org/en/latest/whats-new/upgradingfrom6to7.html) and found an item that seems to match the issue I have.

<img width="678" alt="image" src="https://user-images.githubusercontent.com/325813/224761101-3737cff6-04b6-491c-ac40-234a07e8fa8b.png">

Both links are broken

<img width="801" alt="image" src="https://user-images.githubusercontent.com/325813/224761264-f2168ff8-38d6-4318-b5bd-c09754bc1147.png">

## Expected Behavior

I expected to read the docs and fix my build

## Dependency Versions

Autofac: 7

tillig:
Turns out I also had to remove the trailing underscores, but I verified the links work now.

PR #170, opened 2023-03-13
Fix links

steveoh:
unsure if this is the fix or not but the actual page is missing the .html extension

refs #169


tillig (approved):
This should do it. The HTML extension is generated because, if we generate PDF or some other output format, these serve as "references" not actual HTML doc links, and get translated accordingly.

PR #171, opened 2023-03-25
Rename undeclared variable

InteXX:
The variable `scope` is never declared; it should instead be named `child2`, as `child2SubScope` will be a nested scope under the `child2` scope.


tillig (approved):
Looks great! Thanks!

InteXX:
> Looks great! Thanks!

...and thank you as well.

PR #176, opened 2023-09-29
Refer to TypedParameter extension method

aydjay:
This is a more elegant method of specifying typed parameters


tillig (approved):
I like it, thanks!

PR #178, opened 2023-10-18
Add dependency for sphinx-rtd-theme and explictly specify theme in conf.py

### LINKED ISSUE #177: The layout seems completely broken.
Some time ago, the site looked like this:
https://web.archive.org/web/20220412193519/https://autofac.readthedocs.io/en/latest/

Rather easy to navigate and visually pleasing.

Now, it's nowhere near that. The theme has changed to "classic", and it's not good. I doubt it was intentional.

alistairjevans:
Fixes #177 

Thanks for the heads up @gliljas


tillig (approved):
🚔

PR #204, opened 2026-09-01
Document minimal hosting and ASP.NET Core parameter binding

### LINKED ISSUE #175: Please add good getting started for single program template .NET 6 project
ITNOA

Please add good example that fit to .NET 6 Web API new template project

as you can see https://stackoverflow.com/questions/69754985/adding-autofac-to-net-core-6-0-using-the-new-single-file-template some people have same problem

### LINKED ISSUE #179: Extend ASP.NET Core Integration documentation to reflect the parameter binding changes made in .NET 7
In .NET 7, when `IServiceProviderIsService.IsService` was introduced, the ASP.NET Core framework attempts to detect if something is a service, and if we indicate it is, that will be used as the source of binding, as opposed to the `[FromBody]` binding, which was the default before.

In Autofac rules, all collections are valid services (but with an empty set if no registrations have been made), so from .NET 7 on, ASP.NET Core will inject empty collections from Autofac instead of binding the body content.

We should add to our docs for ASP.NET Integration to detail this point and indicate that users should be explicit with `[FromBody]` where appropriate to prevent automatic service injection.

See https://github.com/autofac/Autofac.Extensions.DependencyInjection/issues/114

tillig:
Fixes #175. Fixes #179.

## Minimal hosting (#175)

The page only ever showed the `Startup` class style, so the shape the current templates actually produce was undocumented — which is what #175 asked for. Minimal hosting now leads; `Startup` is kept after it, since plenty of applications still use it.

The rule people trip over — **you cannot return an `IServiceProvider` from `ConfigureServices`, and cannot add the factory to the service collection** — moved up into the Quick Start, because it applies to both styles and was previously buried in one of them.

## Parameter binding (#179)

This is the one worth reading. Since .NET 7, ASP.NET Core asks `IServiceProviderIsService` whether a parameter is a service, to decide between injecting it and binding it from the request body. **Autofac answers yes for every collection type**, because collections always resolve and simply come back empty when nothing is registered.

So a collection parameter silently stops binding from the body and starts arriving empty:

```csharp
// On .NET 6 this bound from the request body.
// On .NET 7+ it arrives empty.
[HttpPost]
public IActionResult Post(IReadOnlyCollection<WeatherForecast> forecasts)
```

Nothing throws, which is exactly why it earns a warning directive. The fix is an explicit `[FromBody]`. The section also explains *why* Autofac won't special-case empty collections — it would make the answer depend on registration order, break the case where you do want an empty set injected, and contradict documented collection behavior — so the guidance doesn't read like an unfixed oversight.

Verified against `AutofacServiceProvider.IsService`, which is `ComponentRegistry.IsRegistered(new TypedService(serviceType))`, and the exact collection set in `InternalTypeExtensions.IsGenericListOrCollectionInterfaceType` plus `CollectionRegistrationSource`.

**That turned up a second gap.** `resolve/relationships.rst` listed `IEnumerable<B>`, `IList<B>` and `ICollection<B>` but never mentioned `IReadOnlyCollection<B>`, `IReadOnlyList<B>` or arrays — precisely the types reported in the issue. Since the new section points readers there, it's fixed here too. The heading is deliberately left alone: an upstream `Autofac.Extensions.DependencyInjection` issue links that exact anchor.

## Keyed services

`AddKeyedSingleton`, `[FromKeyedServices]` and `KeyedService.AnyKey` all work through Autofac, cross-referenced to core Autofac's own keyed service support. Confirmed against `KeyedServiceTests` and `FromKeyedServicesAttributeExtensions`.

## Code that could not have compiled

The child lifetime scope sample had four defects: a variable named `builder` where `containerBuilder` was declared, constructors named `Startup` inside `StartupOne` and `StartupTwo`, a missing statement terminator on `Task.WhenAll`, and stray indentation. These were the last of the four class/constructor mismatches I catalogued during the structural pass.

## Obsolete APIs and dead links

`app.UseMvc()` becomes routing/endpoint calls, `loggerFactory.AddConsole`/`AddDebug` are gone (logging is configured on the host), and `IHostingEnvironment` becomes `IWebHostEnvironment`.

All **ten** `docs.asp.net` links are gone. The nine-link "helpful links" list collapsed to a single pointer at Microsoft's DI documentation — a list of deep links into another product's docs is what rotted in the first place. Also dropped a 2016 blog post about RC2 behavior and its comment link, and the archived `aspnet/DependencyInjection` link now points into `dotnet/runtime`.

## Coordination note

Your other session committed `feature/link-new-examples` locally, which also touches this file's `Example` section. **I deliberately left that section untouched here** so the two don't collide — that change belongs to its PR, and mine applies cleanly either way round.

## Verification

`sphinx -b html -W --keep-going`, `doc8` and `pre-commit` clean. Rendered HTML read for both new sections, and the page heading tree checked.

---

## Follow-up: nested inline markup

The other session spotted a real bug here while my files were briefly on its branch — thanks for passing it along. The Quick Start callout was:

```rst
**You cannot return an ``IServiceProvider`` from ``ConfigureServices``, ...**
```

RST doesn't nest inline markup, so those double backticks rendered as literal backtick characters inside the bold:

```html
<strong>You cannot return an ``IServiceProvider`` from ``ConfigureServices``, ...</strong>
```

`sphinx -W` can't catch this — the markup is perfectly valid, just not what was meant — which is why it survived the build being clean. Fixed by closing the bold around each literal, the same pattern the `[FromBody]` callout further down this page already used. It now renders as `<strong>` plus `<code>`, verified in the built HTML.

I scanned every page for the whole bug class, in both nesting directions (literal-inside-bold, literal-inside-italic, bold-inside-literal). Exactly two instances existed: this one and `integration/servicefabric.rst:115`, which had the identical defect and rendered wrong for the same reason.

**That second fix is included here** even though it predates this branch and isn't ASP.NET Core. It's one line, it's the same defect, and leaving a known-broken render in place purely to keep the diff on-topic seemed like the wrong trade. Easy to split out if you'd rather.

### DIFF
--- a/docs/integration/aspnetcore.rst
+++ b/docs/integration/aspnetcore.rst
@@ -23,7 +23,7 @@ Quick Start
 
 The ``IServiceProvider`` gets created for you, so there's nothing to do but *register things*. Autofac calls ``Populate`` on your behalf to move what's in the service collection into the container.
 
-**You cannot return an ``IServiceProvider`` from ``ConfigureServices``, and you cannot add the factory to the service collection.** It has to be handed to the host builder directly. That is the one rule that trips people up, whichever hosting style you use.
+**You cannot return an** ``IServiceProvider`` **from** ``ConfigureServices``\ **, and you cannot add the factory to the service collection.** It has to be handed to the host builder directly. That is the one rule that trips people up, whichever hosting style you use.
 
 Minimal Hosting
 ---------------
--- a/docs/integration/aspnetcore.rst
+++ b/docs/integration/aspnetcore.rst
@@ -490,4 +490,10 @@ Using a child lifetime scope as the root is not compatible with multitenant supp
 Example
 =======
 
-There is an example project showing ASP.NET Core integration `in the Autofac examples repository <https://github.com/autofac/Examples/tree/main/src/AspNetCoreExample>`_.
+The Autofac examples repository has a runnable project for each style described above:
+
+* `AspNetCoreExample <https://github.com/autofac/Examples/tree/main/src/AspNetCoreExample>`_ - a ``Startup`` class with ``ConfigureContainer``
+* `AspNetCoreNoStartupExample <https://github.com/autofac/Examples/tree/main/src/AspNetCoreNoStartupExample>`_ - the same application under minimal hosting, with no ``Startup`` class
+* `AspNetCoreChildLifetimeScope <https://github.com/autofac/Examples/tree/main/src/AspNetCoreChildLifetimeScope>`_ - two hosts sharing one container, each rooted in its own child lifetime scope
+
+Multitenancy has its own example, `MultitenantExample.AspNetCore <https://github.com/autofac/Examples/tree/main/src/MultitenantExample.AspNetCore>`_, covered on the :doc:`multitenant applications <../advanced/multitenant>` page.
--- a/docs/integration/aspnetcore.rst
+++ b/docs/integration/aspnetcore.rst
@@ -2,7 +2,7 @@
 ASP.NET Core
 ============
 
-ASP.NET Core (previously ASP.NET 5) changes the way dependency injection frameworks have previously integrated into ASP.NET execution. Previously, each functionality - MVC, Web API, etc. - had its own "dependency resolver" mechanism and just slightly different ways to hook in. ASP.NET Core introduces a `conforming container <https://blog.ploeh.dk/2014/05/19/conforming-container/>`__ mechanism via `Microsoft.Extensions.DependencyInjection <https://github.com/aspnet/DependencyInjection>`_, including a unified notion of request lifetime scope, service registration, and so forth.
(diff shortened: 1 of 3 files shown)


tillig:
Added the example link fixup here as agreed, in `cf3f593` — one commit touching only the `Example` section of `docs/integration/aspnetcore.rst`.

The section listed only `AspNetCoreExample`, and `AspNetCoreNoStartupExample` and `AspNetCoreChildLifetimeScope` had no link from anywhere in the documentation. Since this PR splits the page into minimal hosting and `Startup` class styles, both of those now have a runnable counterpart a reader can actually find. `MultitenantExample.AspNetCore` gets a pointer to the multitenant page, where it is already covered, rather than a duplicate entry.

With this in, every example project in the Examples repository is linked from the page that documents it. The only two without links are `ConfigurationExampleInterface` and `ConfigurationExamplePlugin`, which are supporting libraries for `ConfigurationExample` and correctly have no page of their own.

Verified by simulating the merge — main with this branch's versions of its two files — and checking every `Examples/tree/<branch>/src/<dir>` reference against a local clone: all on `main`, all pointing at a directory that exists. `pre-commit` including `doc8` passes. Also confirmed `6c6b800` fixed the nested inline literal I flagged earlier.

PR #205, opened 2026-09-01
Link example projects from their documentation pages

tillig:
Part of the Examples modernization tracked in autofac/Examples#32.

## Proposed Changes

**Every example is now reachable from the page that documents it.** That was already true in the other direction — each example README links to its documentation page — but eight examples had no inbound link.

New `Example` sections on the interceptors, pooled instances, tracing, .NET Core, JSON/XML configuration, and metadata pages. The multitenant page gains the ASP.NET Core sample. The resolve pipeline examples now point at the runnable project, calling out the one case that does not work the obvious way: middleware substituting parameters has to be added with `ConfigurePipeline`, because `ParameterSelection` is not a phase a service pipeline accepts.

That leaves only `ConfigurationExampleInterface` and `ConfigurationExamplePlugin` unlinked, which is correct — they are supporting libraries for `ConfigurationExample`, not examples in their own right.

## Verification

`pre-commit` passes, including `doc8` and markdownlint. All links into the Examples repository resolve to a directory that exists and use `/tree/main`, checked against a local clone rather than by eye. Section underlines are at least as long as their titles in every changed file.

## Note

An earlier revision of this branch also contained the `docs/integration/aspnetcore.rst` and `docs/resolve/relationships.rst` changes that belong to #204. That was my mistake — I picked them up out of a shared working tree and committed them here. The branch has been rewritten to drop them, so this PR is now only the eight files above and does not overlap #204.


tillig:
Rewrote this branch to drop two files that belong to #204. An earlier revision had picked up `docs/integration/aspnetcore.rst` and `docs/resolve/relationships.rst` from a shared working tree; they are gone from here and #204 is unaffected.

One thing found while those files were briefly on this branch, worth fixing in #204: the callout near the top of the Quick Start reads

    **You cannot return an ``IServiceProvider`` from ``ConfigureServices``, and you cannot add the factory to the service collection.**

RST does not allow inline markup to nest, so the double backticks inside the bold will render literally rather than as code. Either close the bold around the literals, which is what ``[FromBody]`` does further down that same page, or reword so the bold contains no literal.

The same pattern already exists on `integration/servicefabric.rst` line 115 and renders wrong there too, unrelated to either PR.

PR #211, opened 2026-09-02
Correct the release process and refresh the meta pages

tillig:
Phase 3g. The sharpest defect left in the docs, plus the meta-page cleanup.

## The release process was documented backwards

`owners.rst` told a new owner to:

> Merge `develop` into `master` and tag the repo with `vX.Y.Z` […] After the CI build is complete and pushed to MyGet, push the package from MyGet to NuGet as the final release. […] download the published package manually from NuGet […] Upload the .nupkg

**A merge is the one operation that must not happen.** `develop` and `main` have to end up on the *exact same commit hash*, because the release ships the commit that was built and validated on `develop`. A `--no-ff` merge creates a commit that exists only on `main` — never built on `develop`, changing the released hash, and leaving `develop` permanently one behind. The guide was instructing owners to cause precisely the drift the real procedure exists to prevent.

The section now documents the fast-forward-only procedure, states which operations are forbidden and why, and includes the recovery if a merge commit does land (fast-forward `develop` up to `main`, which moves a pointer rather than creating a commit).

## Publishing, checked against the CI rather than rewritten from memory

From the shared `ci.yml`, the `publish` job runs on `develop` pushes and on `v*` tags:

| Trigger | GitHub Packages | NuGet.org |
|---|---|---|
| push to `develop` | yes | no |
| `vX.Y.Z` tag | yes | **yes** |

So **the tag push is what releases** — not the push of `main`. MyGet is gone, and there's no manual download-and-upload step: core Autofac stopped attaching packages to releases in 8.3.0, which also retires the old rationale about NuGet being filtered in some regions.

## Support channels now agree with each other

Four pages disagreed. `owners.rst`, `support.rst`, `best-practices` and `getting-started` now all say StackOverflow, the Autofac group, and GitHub issues. Gitter and Twitter dropped, commercial support section removed, and the four Google Group links — which were in **three different URL formats** — are all `groups.google.com/g/autofac`.

**With these gone, the docs contain no dead hosts anywhere.** That closes out the link work started in #198.

## Two stale recommendations

- **SlimTune** was hosted on Google Code, shut down in 2016. The profiler list is now three live options.
- **"up to 10x faster"** for lambda registration is a precise number nobody can substantiate against a current release, and it predates both the 6.x pipeline rework and the 9.1 performance pass. Replaced with what is actually true — reflection is skipped, the gain depends on the graph — plus a pointer to Autofac's benchmark suite for anyone who wants to measure their own workload. I considered running the `LambdaResolveBenchmark` to produce a current figure, but a laptop BenchmarkDotNet run isn't a number worth publishing.
- The CodeProject tutorial link is dropped: it predates Autofac 3 and shows APIs that no longer exist.

## CONTRIBUTING.md

Five references to a `master` branch that doesn't exist; **eslint listed as a pre-commit hook** when the hooks are `check-json`, `check-yaml`, `check-merge-conflict`, `end-of-file-fixer`, `trailing-whitespace`, `markdownlint`, `doc8`, `json-sort` (eslint runs via `npm run lint`); and a Windows-only `.venv\Scripts\Activate.ps1` under a comment claiming it was cross-platform. It now also shows the docs build CI runs, since that's the check which catches a broken cross-reference the linters can't see.

The rotting deep-link list in References collapsed to three durable targets, and `docs.readthedocs.org` / `docutils.sourceforge.net` moved to their current `.io` domains.

## Verification

The release procedure is transcribed from the project's own documented process; publishing behavior from `autofac/.github`'s `ci.yml`; the hook list from `.pre-commit-config.yaml`.

`sphinx -W`, `doc8`, `pre-commit`, `npm run lint` all clean. Both custom scans clean.

Two things the tooling caught on me during this, worth noting since they're the reason those checks exist: I wrote **nested inline markup** in three places in the new release section (literals inside bold), caught by the scan; and unifying a StackOverflow URL created a **duplicate explicit target name**, caught by `doc8`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 User documentation is viewable at <https://docs.autofac.org> (a CNAME to <https://autofac.readthedocs.io>). It is stored in the `/docs` folder in this source repo.
 
-We also have API docs at <https://autofac.org/apidoc/>. This documentation is what gets put in the <https://github.com/autofac/autofac.github.com/tree/master/apidoc> folder.
+We also have API docs at <https://autofac.org/apidoc/>. This documentation is what gets put in the <https://github.com/autofac/autofac.github.com/tree/main/apidoc> folder.
 
 - [Checkout and Setup](#checkout-and-setup)
 - [Validating Changes](#validating-changes)
@@ -30,8 +30,8 @@ After cloning, set up the tools and dependencies.
 # Set up your Python virtual environment
 python3 -m venv .venv
 
-# Activate the virtual environment (cross-platform)
-.venv\Scripts\Activate.ps1
+# Activate the virtual environment
+.venv/bin/Activate.ps1   # .venv\Scripts\Activate.ps1 on Windows
 
 # Restore dependencies
 npm install
@@ -42,7 +42,7 @@ dotnet tool restore
 pre-commit install
 
 # At times you may need to FORCE PULL tags because the most recent doc version
-# tag follows the `master` branch.
+# tag follows the `main` branch.
 git pull --tags --force
 
 # When you're done, deactivate your virtual environment.
@@ -53,7 +53,6 @@ deactivate
 
 This repository uses `pre-commit` hooks to automatically validate code quality. These hooks run automatically when you commit, and include:
 
-- **eslint**: Lints JavaScript files
 - **markdownlint**: Checks Markdown formatting in `.md` files
 - **doc8**: Validates reStructuredText (`.rst`) files in the `docs/` folder for formatting and style issues
 - **General checks**: YAML/JSON validation, trailing whitespace, merge conflict markers, etc.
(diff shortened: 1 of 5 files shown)


tillig:
Follow-up: the release-process warning contradicted the procedure above it.

It read **"Never merge `develop` into `main`"**, three steps after the procedure runs exactly that merge with `--ff-only`. The rule was always about the merge *commit*, not the merge *command* — so it now says that:

> **Never create a merge commit between the two branches, in either direction.** `--ff-only` is the only merge that should ever run against them; if it refuses, stop and reconcile rather than dropping the flag.

That also resolves the recovery paragraph, which fast-forwards `develop` up to `main` and was therefore itself an instance of the merge the old wording forbade.

The instruction *not* to merge back into `develop` is kept, because standard Gitflow tells you to and a reader following it would be wrong here — but now with its reason: after a fast-forward both branches are already on the same commit, so there is nothing to merge back.

Fixed in both copies so the relocation doesn't carry the flaw forward.

